PR #39, opened 2020-02-24
Allow  to manually assign ip addresses to the nat gateway

### LINKED ISSUE #1: [GCP] Expose configuration for GCP Cloud NAT manual external IP address
As mentioned in the google documentation https://cloud.google.com/nat/docs/using-nat#update_nat, there is a way to set a reserved static IP address for Cloud NAT. Currently gardener doesnt' have a configuration available for setting it to be manual instead of automatic. The reason for requesting this configuration is that we have an use case which requires us to whitelist our clusters' public IP addresses in different places, that don't allow access unless the IP address is officially whitelisted.

DockToFuture:
This PR allows to manually assign ip addresses to the nat gateway. A list of ip addresses can be specified in the ```shoot.yaml```. See the following configuration, where two ip addresses are added to the nategateway. The ip addresses must be specified by their name. 
```
spec:
  provider
    infrastructureConfig
      cloudNAT:
        minPortsPerVM: 2048
        natIPNames:
        - name: manualnat1
        - name: manualnat2
```
After reconciliaton the status field will be updated with the values of the respective ip addresses.
```
status
    providerStatus:
      kind: InfrastructureStatus
      networks:
        natIPs:
        - ip: 34.89.187.69
        - ip: 35.246.255.135
```

Furthermore a validation has to be added where we assure that the number the number of ip addresses specified is sufficient for the number of vms and minimum ports per instance value which is set. A sample for the calculation can be found here:  https://cloud.google.com/nat/docs/ports-and-addresses#port-reservation-examples
The validation and calculation will be added in a separate PR.

Additional:

Currently gcp only allows to add premium ip addresses to the nat gateway. There refer to using them when reserving static ip addresses.

Fixes #1 

**Release note**:

```improvement operator
It is now possible to assign manual ip addresses to the nat gateway by specifying them in `spec.provider.infrastructureConfig.networks.cloudNAT.natIPNames` section in the `shoot.yaml` file. If the section is not provided then the nat gateway will automatically allocate ip addresses.
```
```improvement operator
Please consider that the #IPs for the nat gateway is linked to the #Nodes and portsPerVM. See https://cloud.google.com/nat/docs/ports-and-addresses#port-reservation-examples for more details.
```


zanetworker:
@DockToFuture can you add another release note with regards to the IP address calculation? https://cloud.google.com/nat/docs/ports-and-addresses#port-reservation-examples 

We should probably decide, whether we want to delegate the validation to the users, or implement logic for validating # of nodes, portsPerVM, and # of IPs.

pkg/apis/gcp/v1alpha1/types_infrastructure.go
```diff
@@ -111,6 +114,26 @@ type CloudNAT struct {
 	// The default value is 2048 ports.
 	// +optional
 	MinPortsPerVM *int32 `json:"minPortsPerVM,omitempty"`
+	// NatIPNames is a list of all user provided external premium ips which can be used by the nat gateway
+	// +optional
+	NatIPNames *[]NatIPName `json:"natIPNames,omitempty"`
+	// NatIPs is a list of all user provided external premium ips which can be used by the nat gateway
+	// +optional
+	NatIPs *[]NatIP `json:"natIPs,omitempty"`
```
rfranzke: Can we harmonize `NatIPNames` and `NatIPs`?

```golang
NatIPs []NatIP `json:"natIPs,omitempty"`
```

```golang
type NatIP struct {
  IP *string `json:"ip,omitempty"`
  Name *string `json:"name,omitempty"`
}
```

Only one (ip or name) can be set at a time. WDYT?

pkg/apis/gcp/v1alpha1/types_infrastructure.go
```diff
@@ -70,6 +70,9 @@ type NetworkStatus struct {
 
 	// Subnets are the subnets that have been created.
 	Subnets []Subnet `json:"subnets"`
+
+	// CloudNAT contains configation about the the CloudNAT resource
```
rfranzke:
```suggestion
	// CloudNAT contains configuration about the the CloudNAT resource
```

example/30-infrastructure.yaml
```diff
@@ -59,6 +59,9 @@ spec:
     # internal: 10.243.0.0/19
     # cloudNAT:
     #   minPortsPerVM: 2048
+    #   natIPNames:
```
rfranzke: This lacks the example for IPs instead of names.

docs/usage-as-end-user.md
```diff
@@ -62,6 +65,8 @@ The `networks.internal` section is optional and can describe a CIDR for a subnet
 
 The `networks.cloudNAT.minPortsPerVM` is optional and is used to define the [minimum number of ports allocated to a VM for the CloudNAT](https://cloud.google.com/nat/docs/overview#number_of_nat_ports_and_connections)
 
+The `networks.cloudNAT.natIPNames` is optional and is used to specify the names of the manual ip addresses which should be used by the nat gateway
```
rfranzke: Same here, this document lacks the example for IPs instead of names.

charts/internal/gcp-infra/templates/main.tf
```diff
@@ -55,7 +55,11 @@ resource "google_compute_router_nat" "nat" {
   router =  "${google_compute_router.router.name}"
   {{ end -}}
   region                             = "{{ required "google.region is required" .Values.google.region }}"
-  nat_ip_allocate_option             = "AUTO_ONLY"
+  nat_ip_allocate_option             = {{ if .Values.networks.cloudNAT.natIPNames }}"MANUAL_ONLY"{{ else }}"AUTO_ONLY"{{ end }}
```
rfranzke: Can you specify the new Helm chart values in the `values.yaml` to show what is possible to be configured?

charts/internal/gcp-infra/templates/main.tf
```diff
@@ -55,7 +55,11 @@ resource "google_compute_router_nat" "nat" {
...
+  {{ if .Values.networks.cloudNAT.natIPs -}}
+  nat_ips                = {{ .Values.networks.cloudNAT.natIPs }}
```
rfranzke: Quotes?

pkg/apis/gcp/validation/infrastructure.go
```diff
@@ -84,6 +84,10 @@ func ValidateInfrastructureConfig(infra *apisgcp.InfrastructureConfig, nodesCIDR
 		allErrs = append(allErrs, field.Invalid(networksPath.Child("vpc", "cloudRouter"), infra.Networks.VPC.CloudRouter, "cloud router can not be configured when the VPC name is not specified"))
 	}
 
+	if infra.Networks.CloudNAT != nil && infra.Networks.CloudNAT.NatIPs != nil {
```
rfranzke: What's the reason for this limitation? Let's say I want to attach two NAT IPs, one by name and one by IP. Why is this forbidden/not possible?

DockToFuture: The data source requires a unique name and we need it to extract the ip address for the status and get the selflink for the nat gateway. If there is a better way I'm happy to adapt it.

example/30-infrastructure.yaml
DockToFuture: We only allow names here, because we require the unique name for the data source.

docs/usage-as-end-user.md
DockToFuture: See comment above.

charts/internal/gcp-infra/templates/main.tf
DockToFuture: Yes, we can do that.

pkg/apis/gcp/validation/infrastructure.go
rfranzke: I don't understand. You say this `<NAME>` here must be unique?

```terraform
data "google_compute_address" "<NAME>" {
```

What's the issue with allowing IPs then?

DockToFuture:
```terraform
data "google_compute_address" "Name" {
  name = "<NAME>"
}
```

The ```<NAME>``` here must be unique, how will I get the name for the ip? We could implement it via the sdk but not sure, if that will suit the use case.

rfranzke: OK, I see, understood, but why do we have then https://github.com/gardener/gardener-extension-provider-gcp/pull/39/files#diff-50ba7c5f1771b36c61263800bc8ea823R122 in the first place?

DockToFuture: For convenience, when you want to debug something or look them up, you don't have to go each and every time to the google cloud console. If you think it's better to remove it from the status we can also do that. But I think it will make things easier for us.

rfranzke: It's fine to have it in the status, but as of now an end-user can configure `.networks.cloudNAT.natIPs` with should not be possible as it cannot be used/nothing is implemented.

DockToFuture: https://github.com/gardener/gardener-extension-provider-gcp/pull/39/files#diff-4863b24e73b205b0206ea0c2b7e55d86R88 isn't this doing the validation?

DockToFuture: Or do you mean something else?

rfranzke: My point is that it shouldn't be in the API in the first place if it's not supported.

DockToFuture: Got it, so what do you suggest? Removing it from CloudNat and adding it only to the status or removing it at all?

rfranzke: I‘m suggesting to remove it from the spec. If you consider it helpful having it in the status then I‘m fine with it.

charts/internal/gcp-infra/templates/main.tf
rfranzke: What about this comment?

charts/internal/gcp-infra/values.yaml
```diff
@@ -16,6 +16,9 @@ clusterName: test-namespace
 networks:
   cloudNAT:
     minPortsPerVM: 2048
+#   natIPs: ["${data.google_compute_address.manualnat1.self_link}","${data.google_compute_address.manualnat2.self_link}"]
+#   natIPNames: ["manualnat1", "manualnat2"]
+#   natIPOutput:   "${data.google_compute_address.manualnat1.address},${data.google_compute_address.manualnat2.address}"
```
rfranzke: Wait a sec. Why does a user has to specify this? Same question for `natIPs`.
Imo, `natIPNames` is the only value that should be specified by a user. The other values can be computed on the fly and are an implementation detail, or?

charts/internal/gcp-infra/templates/main.tf
```diff
@@ -176,6 +186,12 @@ output "{{ .Values.outputKeys.cloudNAT }}" {
 }
 {{- end }}
 
+{{ if .Values.networks.cloudNAT.natIPOutput -}}
+output "nat_ips" {
```
rfranzke: You are already injecting the output key in the values, so please use it here.

pkg/apis/gcp/types_infrastructure.go
```diff
@@ -100,11 +104,28 @@ type CloudRouter struct {
...
 
-// CloudNAT contains information about the the CloudNAT configuration
+// CloudNAT contains configuration about the the CloudNAT resource
 type CloudNAT struct {
 	// MinPortsPerVM is the minimum number of ports allocated to a VM in the NAT config.
 	// The default value is 2048 ports.
 	MinPortsPerVM *int32
+	// NatIPNames is a list of all names of user provided external premium ips which can be used by the nat gateway
+	// +optional
+	NatIPNames *[]NatIPName
```
rfranzke:
```suggestion
	NatIPNames []NatIPName
```

pkg/apis/gcp/types_infrastructure.go
```diff
@@ -100,11 +104,28 @@ type CloudRouter struct {
...
 }
 
-// CloudNAT contains information about the the CloudNAT configuration
+// CloudNAT contains configuration about the the CloudNAT resource
 type CloudNAT struct {
 	// MinPortsPerVM is the minimum number of ports allocated to a VM in the NAT config.
 	// The default value is 2048 ports.
 	MinPortsPerVM *int32
+	// NatIPNames is a list of all names of user provided external premium ips which can be used by the nat gateway
+	// +optional
```
rfranzke: not needed in the internal API version - remove

pkg/apis/gcp/types_infrastructure.go
```diff
@@ -66,6 +66,10 @@ type NetworkStatus struct {
 
 	// Subnets are the subnets that have been created.
 	Subnets []Subnet
+
+	// NatIPs is a list of all user provided external premium ips which can be used by the nat gateway
+	// +optional
```
rfranzke: not needed in the internal API version - remove

pkg/apis/gcp/types_infrastructure.go
```diff
@@ -100,11 +104,28 @@ type CloudRouter struct {
...
 	MinPortsPerVM *int32
+	// NatIPNames is a list of all names of user provided external premium ips which can be used by the nat gateway
+	// +optional
+	NatIPNames *[]NatIPName
+}
+
+// NatIP is a user provided external ip which can be used by the nat gateway
+type NatIP struct {
+	// Name of the external premium ip address which is used in gcp
+	// +optional
```
rfranzke: not needed in the internal API version - remove

pkg/apis/gcp/types_infrastructure.go
```diff
@@ -100,11 +104,28 @@ type CloudRouter struct {
...
+	IP *string
```
rfranzke:
```suggestion
	IP string
```

pkg/apis/gcp/types_infrastructure.go
```diff
@@ -100,11 +104,28 @@ type CloudRouter struct {
...
+}
+
+// NatIPName is the name of a user provided external ip address which can be used by the nat gateway
+type NatIPName struct {
+	// Name of the external premium ip address which is used in gcp
+	// +optional
```
rfranzke: not needed in the internal API version - remove

pkg/apis/gcp/types_infrastructure.go
```diff
@@ -100,11 +104,28 @@ type CloudRouter struct {
...
+	Name *string
```
rfranzke:
```suggestion
	Name string
```

pkg/apis/gcp/v1alpha1/types_infrastructure.go
```diff
@@ -70,6 +70,10 @@ type NetworkStatus struct {
 
 	// Subnets are the subnets that have been created.
 	Subnets []Subnet `json:"subnets"`
+
+	// NatIPs is a list of all user provided external premium ips which can be used by the nat gateway
+	// +optional
+	NatIPs *[]NatIP `json:"natIPs,omitempty"`
```
rfranzke:
```suggestion
	NatIPs []NatIP `json:"natIPs,omitempty"`
```

pkg/apis/gcp/types_infrastructure.go
```diff
@@ -66,6 +66,10 @@ type NetworkStatus struct {
 
 	// Subnets are the subnets that have been created.
 	Subnets []Subnet
+
+	// NatIPs is a list of all user provided external premium ips which can be used by the nat gateway
+	// +optional
+	NatIPs *[]NatIP
```
rfranzke:
```suggestion
	NatIPs []NatIP
```

pkg/apis/gcp/v1alpha1/types_infrastructure.go
```diff
@@ -105,12 +109,29 @@ type CloudRouter struct {
...
-// CloudNAT contains information about the the CloudNAT configuration
+// CloudNAT contains configuration about the the CloudNAT resource
 type CloudNAT struct {
 	// MinPortsPerVM is the minimum number of ports allocated to a VM in the NAT config.
 	// The default value is 2048 ports.
 	// +optional
 	MinPortsPerVM *int32 `json:"minPortsPerVM,omitempty"`
+	// NatIPNames is a list of all user provided external premium ips which can be used by the nat gateway
+	// +optional
+	NatIPNames *[]NatIPName `json:"natIPNames,omitempty"`
```
rfranzke:
```suggestion
	NatIPNames []NatIPName `json:"natIPNames,omitempty"`
```

pkg/apis/gcp/v1alpha1/types_infrastructure.go
```diff
@@ -105,12 +109,29 @@ type CloudRouter struct {
...
+	// Name of the external premium ip address which is used in gcp
+	// +optional
+	IP *string `json:"ip,omitempty"`
+}
+
+// NatIPName is the name of a user provided external ip address which can be used by the nat gateway
+type NatIPName struct {
+	// Name of the external premium ip address which is used in gcp
+	// +optional
+	Name *string `json:"name,omitempty"`
```
rfranzke:
```suggestion
	Name string `json:"name"`
```

pkg/apis/gcp/v1alpha1/types_infrastructure.go
```diff
@@ -105,12 +109,29 @@ type CloudRouter struct {
...
+type NatIP struct {
+	// Name of the external premium ip address which is used in gcp
+	// +optional
+	IP *string `json:"ip,omitempty"`
+}
+
+// NatIPName is the name of a user provided external ip address which can be used by the nat gateway
+type NatIPName struct {
+	// Name of the external premium ip address which is used in gcp
+	// +optional
```
rfranzke: should not be optional - remove

pkg/apis/gcp/v1alpha1/types_infrastructure.go
```diff
@@ -105,12 +109,29 @@ type CloudRouter struct {
...
 	MinPortsPerVM *int32 `json:"minPortsPerVM,omitempty"`
+	// NatIPNames is a list of all user provided external premium ips which can be used by the nat gateway
+	// +optional
+	NatIPNames *[]NatIPName `json:"natIPNames,omitempty"`
+}
+
+// NatIP is a user provided external ip which can be used by the nat gateway
+type NatIP struct {
+	// Name of the external premium ip address which is used in gcp
+	// +optional
```
rfranzke: should not be optional - remove

pkg/apis/gcp/v1alpha1/types_infrastructure.go
```diff
@@ -105,12 +109,29 @@ type CloudRouter struct {
...
+	IP *string `json:"ip,omitempty"`
```
rfranzke:
```suggestion
	IP string `json:"ip"`
```

pkg/internal/infrastructure/terraform.go
```diff
@@ -135,6 +139,29 @@ func ComputeTerraformerChartValues(
 		},
 	}
 
+	if config.Networks.CloudNAT != nil && config.Networks.CloudNAT.NatIPNames != nil {
+		cN := make(map[string]interface{})
+		if config.Networks.CloudNAT.MinPortsPerVM != nil {
```
rfranzke: Why is this code here? The min ports per VM are already set in line 88.

pkg/internal/infrastructure/terraform.go
```diff
@@ -135,6 +139,29 @@ func ComputeTerraformerChartValues(
 		},
 	}
 
+	if config.Networks.CloudNAT != nil && config.Networks.CloudNAT.NatIPNames != nil {
```
rfranzke: Why don't we move these computations into the if condition after line 86 where we already check for non-nil `CloudNAT`? Then we could also easier compute the `networks` value and probably don't need the type casting in line 162.

charts/internal/gcp-infra/values.yaml
DockToFuture: They are calculated on the fly, will remove the comments.

pkg/apis/gcp/types_infrastructure.go
DockToFuture: Shouldn't optional paramters be pointer?

pkg/internal/infrastructure/terraform.go
DockToFuture: That's true, but we only want to append those values to the values map if `config.Networks.CloudNAT.NatIPNames != nil` or i'm wrong? Then we have to append it after calculating the values.

charts/internal/gcp-infra/templates/main.tf
DockToFuture: Quotes are not required here

pkg/internal/infrastructure/terraform.go
DockToFuture: Calculated it now before and appended it afterwards.

pkg/internal/infrastructure/terraform.go
```diff
@@ -87,6 +92,20 @@ func ComputeTerraformerChartValues(
 		if config.Networks.CloudNAT.MinPortsPerVM != nil {
 			minPortsPerVM = *config.Networks.CloudNAT.MinPortsPerVM
 		}
+		if config.Networks.CloudNAT.NatIPNames != nil {
+			natIPNames := []string{}
+			nIPs := []string{}
+			natIPOutputList := []string{}
+			for _, v := range config.Networks.CloudNAT.NatIPNames {
+				natIPNames = append(natIPNames, v.Name)
+				nIPs = append(nIPs, fmt.Sprintf("\"${data.google_compute_address.%s.self_link}\"", v.Name))
```
rfranzke: Again, why are you doing such things here in the Go code? IMO **only the NAT IP names should be handed over to the TF Helm chart**. Then you can do such things in the Helm chart itself which has the benefit that all the logic is at one place. WDYT?

pkg/internal/infrastructure/terraform.go
```diff
@@ -87,6 +92,20 @@ func ComputeTerraformerChartValues(
...
+				natIPOutputList = append(natIPOutputList, fmt.Sprintf("${data.google_compute_address.%s.address}", v.Name))
+			}
+			cN["natIPs"] = fmt.Sprintf("[%s]", strings.Join(nIPs[:], ","))
+			cN["natIPNames"] = natIPNames
+			cN["natIPOutput"] = strings.Join(natIPOutputList[:], ",")
+			cN["minPortsPerVM"] = minPortsPerVM
```
rfranzke: Why is this inside the `if config.Networks.CloudNAT.NatIPNames != nil {` condition? It is completely independent from the nat IPs, or? Move it out?

pkg/internal/infrastructure/terraform.go
rfranzke: https://github.com/gardener/gardener-extension-provider-gcp/pull/39#discussion_r384894731

DockToFuture: Because otherwise i have to add a string join function and add it to the template function map, not sure if this is the better way.

pkg/internal/infrastructure/terraform.go
DockToFuture: Why is this independent? If nat ip names are specified those values have to be calculated.

rfranzke: I don't see any specific calculation for the min ports per VM in case the NAT IP names are set. Also, the min ports are already assigned here https://github.com/gardener/gardener-extension-provider-gcp/pull/39/files#diff-4787a75463b112a387ec6927df20e486R144

pkg/internal/infrastructure/terraform.go
rfranzke: I don't get it, sorry. These Terraform expressions belong in the Terraform chart itself. Otherwise, if you would have to change  something in the future, you would have to do it in two places which is hard to maintain and error-prone.

pkg/apis/gcp/v1alpha1/types_infrastructure.go
```diff
@@ -105,12 +109,27 @@ type CloudRouter struct {
...
 	// +optional
 	MinPortsPerVM *int32 `json:"minPortsPerVM,omitempty"`
+	// NatIPNames is a list of all user provided external premium ips which can be used by the nat gateway
+	// +optional
+	NatIPNames []NatIPName `json:"natIPNames,omitempty"`
+}
+
+// NatIP is a user provided external ip which can be used by the nat gateway
+type NatIP struct {
+	// Name of the external premium ip address which is used in gcp
```
zanetworker: this is not the name anymore right? 
```suggestion
	// IP is the external premium IP address used in GCP
```

PR #45, opened 2020-03-11
fix nil pointer when applying machine class chart

mvladev:

fix nil pointer when applying machine class chart

Related https://github.com/gardener/gardener/pull/2037

**Release note**:

```improvement developer
Fix nil pointer when applying machine class chart
```


ialidzhikov:
Nice catch, tahnk you! I hit just the same with the aws provider.
There is a test failure. Could you please check it? https://concourse.ci.gardener.cloud/builds/197551

mvladev:
fixed the broken test

PR #52, opened 2020-03-21
Use the Metadata server to get the hostname of a VM

vpnachev:
Use the Metadata server to get the hostname of a VM.
This is necessary because different distributions are using different means to set the hostname on the node.

Fixes #

**Release note**:

```improvement operator
The hostname is now retrieved from the metadata server.
```

### DIFF
--- a/pkg/webhook/controlplane/ensurer.go
+++ b/pkg/webhook/controlplane/ensurer.go
@@ -215,7 +215,7 @@ func (e *ensurer) EnsureKubeletServiceUnitOptions(ctx context.Context, ectx gene
 	opts = extensionswebhook.EnsureUnitOption(opts, &unit.UnitOption{
 		Section: "Service",
 		Name:    "ExecStartPre",
-		Value:   `/bin/sh -c 'hostnamectl set-hostname $(cat /etc/hostname | cut -d '.' -f 1)'`,
+		Value:   `/bin/sh -c 'hostnamectl set-hostname $(wget -q -O- --header "Metadata-Flavor: Google" http://metadata.google.internal/computeMetadata/v1/instance/hostname | cut -d '.' -f 1)'`,
 	})
 	return opts, nil
 }
--- a/pkg/webhook/controlplane/ensurer_test.go
+++ b/pkg/webhook/controlplane/ensurer_test.go
@@ -373,7 +373,7 @@ var _ = Describe("Ensurer", func() {
 					{
 						Section: "Service",
 						Name:    "ExecStartPre",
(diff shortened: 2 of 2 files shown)


rfranzke:
Thanks @vpnachev - have you verified that this works well for all operating systems without breaking any existing behaviour?

vpnachev:
Yes, it is working with all of the currently supported distributions.

vpnachev:
@zanetworker we were discussing this in private last week, could you provide your opinion as well?

zanetworker:
As discussed, my only concern is having network policies on the cluster that could prevent correct kubelet booting, the chances of this happening though is very small, so I guess it should be okay.

rfranzke:
That's a good point @zanetworker. However, network policies installed via the Kubernetes API do not apply for the kubelet, right? I guess the kubelet needs the metadata service during runtime already for other things. Additional custom-created iptable rules that would also affect kubelet's traffic on the VM is danerous anyways and won't probably work for the mentioned reason. Is this assumption correct?

zanetworker:
Correct. Since network-policy implementation can vary (in how IPtables are installed) from CNI to CNI, I was skeptical about the side effects, ideally, policies are valid for services and pods but for example if we have a pod that is in `hostNetwork` and a policy to block all HTTP traffic from that pod then that could cause a problem. However, I have to say that I have not tried this, neither have I seen this issue happening before.

vpnachev:
According to the CCM [concept](https://kubernetes.io/docs/concepts/architecture/cloud-controller/) the kubelet should not talk to the metadata server any more, but checking the established connections from a v1.15.11 Node

```bash
$ netstat -ntp | grep '169.254.169.254.*ESTABLI'
tcp        0      0 10.242.0.72:62720       169.254.169.254:80      ESTABLISHED 4746/kubelet
tcp        0      0 10.242.0.72:44510       169.254.169.254:80      ESTABLISHED 1498/python
tcp        0      0 10.242.0.72:44512       169.254.169.254:80      ESTABLISHED 1500/python
tcp        0      0 10.242.0.72:43940       169.254.169.254:80      ESTABLISHED 1499/python
``` 

So, if there is no objections by tomorrow evening I will proceed and merge this PR.

zanetworker:
Speaking of, thats exactly what I mean in this issue (https://github.com/gardener/gardener/issues/1953)  with OS processes. 
image

vpnachev:
Yes, the three python process are actually the `google_*` daemons.

rfranzke:
> According to the CCM concept the kubelet should not talk to the metadata server any more

I think this will only be true once we have migrated to CSI (in-progress, see #2).

PR #69, opened 2020-04-21
Add CloudProfile validator

timuthy:
This PR adds a validator for GCP `CloudProfiles`.

The validation logic itself was also changed in a way that it now cross checks the images specified in the cloud profile `spec.machineImages` with the ones given in `spec.providerConfig.machineImages`.

**Release note**:

```improvement operator
Cloud specific settings of GCP `CloudProfiles` are now being validated.
```
```action operator
The `gardener-extension-validator-gcp` Helm chart as well as different assets inside have been renamed to the more general term `gardener-extension-admission-gcp`. Please consider to take corresponding action if you don't use Helm to manage your deployment in the Garden cluster.
```
```action operator
The Docker image `eu.gcr.io/gardener-project/gardener/extensions/validator-gcp` will no longer be maintained as of this release in favor of the successor `eu.gcr.io/gardener-project/gardener/extensions/admission-gcp`. Please consider replacing any references to the image `eu.gcr.io/gardener-project/gardener/extensions/validator-gcp` by `eu.gcr.io/gardener-project/gardener/extensions/admission-gcp`.
```


pkg/webhook/validator/serialization.go
```diff
@@ -48,3 +49,12 @@ func decodeCloudProfileConfig(decoder runtime.Decoder, config *gardencorev1beta1
 
 	return cloudProfileConfig, nil
 }
+
+func decodeCoreCloudProfileConfig(decoder runtime.Decoder, config *core.ProviderConfig) (*gcp.CloudProfileConfig, error) {
```
rfranzke, line 53:
```suggestion
func decodeCloudProfileConfigFromInternalProviderConfig(decoder runtime.Decoder, config *core.ProviderConfig) (*gcp.CloudProfileConfig, error) {
```

pkg/apis/gcp/validation/cloudprofile.go
```diff
@@ -19,37 +19,50 @@ import (
...
-		}
+	return allErrs
+}
 
-		if len(machineImage.Versions) == 0 {
-			allErrs = append(allErrs, field.Required(idxPath.Child("versions"), fmt.Sprintf("must provide at least one version for machine image %q", machineImage.Name)))
-		}
-		for j, version := range machineImage.Versions {
-			jdxPath := idxPath.Child("versions").Index(j)
+func ValidateVersions(versionsConfig []apisgcp.MachineImageVersion, versions []core.ExpirableVersion, fldPath *field.Path) field.ErrorList {
```
rfranzke, line 48: Can you add a doc string for this exported function?

pkg/cmd/options.go
```diff
@@ -53,7 +55,14 @@ func ControllerSwitchOptions() *controllercmd.SwitchOptions {
...
 	return webhookcmd.NewSwitchOptions(
-		webhookcmd.Switch(extensioncontrolplanewebhook.WebhookName, controlplanewebhook.AddToManager),
-		webhookcmd.Switch(extensioncontrolplanewebhook.ExposureWebhookName, controlplaneexposurewebhook.AddToManager),
+		webhookcmd.Switch(extensioncontrolplanewebhook.WebhookName, controlplanewebhook.New),
+		webhookcmd.Switch(extensioncontrolplanewebhook.ExposureWebhookName, controlplaneexposurewebhook.New),
+	)
+}
+
+// GardenWebhookSwitchOptions are the webhookcmd.SwitchOptions for the provider webhooks.
+func GardenWebhookSwitchOptions() *webhookcmd.SwitchOptions {
```
rfranzke, line 64: Would it make sense to put this in a separate `pkg/validator/cmd` package?

timuthy: Hm good question but at least `pkg/validator/cmd` won't be adequate if we add the mutating/defaulting webhook.

rfranzke: What do you mean? I was talking about the `GardenWebhookSwitchOptions` function that, according to my understanding, is only relevant for the validator binary, or? Most things related to the validator binary are already separated in `pkg/validator`, so I was thinking about introducing this function in `pkg/validator/cmd/options.go`.

timuthy: Yes, my thought was a bit forward. In the near future the validator will not be a validator but also a defaulter (both can be disabled by command line flag). Or is your suggestion to have another binary for defaulting provider specific parts of resources in the Garden cluster? My impression is that this would be a bit too undersized.

rfranzke: Ah, understood now. No, I don't suggest a separate binary. Probably the name "validator" doesn't make sense anymore in this case - it should rather be called "admission-controller" or "admission-server" or similarly then. However, this renaming should be done once such defaulting webhooks are introduced, it's not needed now. I still think that for the current state it'd be slightly cleaner to separate these switch options to the `validator` package, so later (when "re-branded" to `admission` or similar) it is already at the right place and can be extended with switch options for defaulters.

pkg/apis/gcp/validation/cloudprofile.go
timuthy: I made it private since it's not required by an external package.

pkg/cmd/options.go
timuthy: Thanks for the feedback. I've moved some packages and renamed some files to the more generic term `admission`, right away (see 80dc11886044f203733fa025a7c8fb0c559953ef). I'm already working on a defaulter but would prefer to open another PR for it for the sake of the reviewers.

pkg/admission/validator/cloudprofile.go
```diff
@@ -0,0 +1,68 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+)
+
+type cloudProfile struct {
+	decoder runtime.Decoder
+}
+
+// NewCloudProfileValidator returns a new instance of a shoot validator.
```
tim-ebert, line 38:
```suggestion
// NewCloudProfileValidator returns a new instance of a cloudprofile validator.
```

pkg/admission/validator/cloudprofile.go
```diff
@@ -0,0 +1,68 @@
...
+
+// InjectScheme injects the given scheme into the validator.
+func (cp *cloudProfile) InjectScheme(scheme *runtime.Scheme) error {
+	cp.decoder = serializer.NewCodecFactory(scheme).UniversalDecoder()
+	return nil
+}
+
+var cpProviderConfigPath = specPath.Child("providerConfig")
+
+// Validate validates the given shoot objects.
```
tim-ebert, line 51:
```suggestion
// Validate validates the given cloudprofile objects.
```

pkg/admission/validator/cloudprofile.go
```diff
@@ -0,0 +1,68 @@
...
+func (cp *cloudProfile) Validate(ctx context.Context, new, old runtime.Object) error {
+	cloudProfile, ok := new.(*core.CloudProfile)
+	if !ok {
+		return fmt.Errorf("wrong object type %q", reflect.TypeOf(new).Name())
```
tim-ebert, line 55: Can you use `%T` here?
```suggestion
		return fmt.Errorf("wrong object type %T", new)
```

pkg/admission/validator/shoot.go
```diff
@@ -19,20 +19,60 @@ import (
...
+func (s *shoot) InjectClient(client client.Client) error {
+	s.client = client
+	return nil
+}
+
+// Validate validates the given shoot objects.
+func (s *shoot) Validate(ctx context.Context, new, old runtime.Object) error {
+	shoot, ok := new.(*core.Shoot)
+	if !ok {
+		return fmt.Errorf("wrong object type %q", reflect.TypeOf(new).Name())
```
tim-ebert, line 63: Same here

pkg/admission/validator/shoot.go
```diff
@@ -19,20 +19,60 @@ import (
...
+	}
+
+	if old != nil {
+		oldShoot, ok := old.(*core.Shoot)
+		if !ok {
+			return fmt.Errorf("wrong object type %q for old object", reflect.TypeOf(old).Name())
```
tim-ebert, line 69: Same here

pkg/admission/validator/webhook.go
```diff
@@ -0,0 +1,45 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package validator
+
+import (
+	"github.com/gardener/gardener-extension-provider-gcp/pkg/gcp"
```
tim-ebert, line 18: Can you sort the imports, please?

timuthy:
Thanks for the valuable feedback. I addressed the comments, PTAL.

pkg/admission/decode.go
```diff
@@ -53,7 +57,18 @@ func decodeInfrastructureConfig(decoder runtime.Decoder, infra *core.ProviderCon
 	return infraConfig, nil
 }
 
-func decodeCloudProfileConfig(decoder runtime.Decoder, config *gardencorev1beta1.ProviderConfig) (*gcp.CloudProfileConfig, error) {
+// DecodeCloudProfileConfigProviderConfig decodes the `CloudProfileConfig` from the given `ProviderConfig`.
+func DecodeCloudProfileConfigProviderConfig(decoder runtime.Decoder, config *core.ProviderConfig) (*gcp.CloudProfileConfig, error) {
```
tim-ebert, line 61: `DecodeCloudProfileConfigProviderConfig` sounds a bit duplicated. WDYT about `DecodeCloudProfileConfig` to keep it consistent with the other funcs?
```suggestion
func DecodeCloudProfileConfig(decoder runtime.Decoder, config *core.ProviderConfig) (*gcp.CloudProfileConfig, error) {
```

timuthy: Sorry, I must have missed that.

pkg/admission/decode.go
```diff
@@ -53,7 +57,18 @@ func decodeInfrastructureConfig(decoder runtime.Decoder, infra *core.ProviderCon
...
+	cloudProfileConfig := &gcp.CloudProfileConfig{}
+	if err := util.Decode(decoder, config.Raw, cloudProfileConfig); err != nil {
+		return nil, err
+	}
+
+	return cloudProfileConfig, nil
+}
+
+// DecodeCloudProfileFromExternalConfig decodes the `CloudProfileConfig` from the given `ProviderConfig`.
+func DecodeCloudProfileFromExternalConfig(decoder runtime.Decoder, config *gardencorev1beta1.ProviderConfig) (*gcp.CloudProfileConfig, error) {
```
tim-ebert, line 71: Sorry for being so nit-picking, but do you want to rename this to `DecodeCloudProfileConfigFromExternalProviderConfig` or something similar? Cause this is rather decoding the provider config than the actual `CloudProfile`.
```suggestion
func DecodeCloudProfileConfigFromExternalProviderConfig(decoder runtime.Decoder, config *gardencorev1beta1.ProviderConfig) (*gcp.CloudProfileConfig, error) {
```

Dockerfile
```diff
@@ -15,8 +15,14 @@ COPY charts /charts
 COPY --from=builder /go/bin/gardener-extension-provider-gcp /gardener-extension-provider-gcp
 ENTRYPOINT ["/gardener-extension-provider-gcp"]
 
-############# gardener-extension-validator-aws
+############# gardener-extension-admission-gcp
+FROM base AS gardener-extension-admission-gcp
+
+COPY --from=builder /go/bin/gardener-extension-admission-gcp /gardener-extension-admission-gcp
+ENTRYPOINT ["/gardener-extension-admission-gcp"]
+
```
rfranzke, line 23: Can you remove everything below this line?

timuthy: I hadn't specified the "old" validator image but then struggled with a failing CI pipeline since the new definition is not effective as long as the PR is not merged. Hence, ddd8e23cd6ddad2403be89f32d5760387aa9e873 adds the old image again and can be reverted once the changes are merged to the master branch. Do you have another idea how to handle this situation, so I can remove the mentioned lines directly?

rfranzke: OK, thanks, seems to be the most pragmatic way. Then let's merge this PR, can you open a new one that reverts the commit again?

timuthy: https://github.com/gardener/gardener-extension-provider-gcp/pull/92

PR #77, opened 2020-04-23
Add missing snapshot-controller component

### LINKED ISSUE #73: Volume snapshotting does not work
**What happened**:

I provisioned a Gardener GCP cluster with CSI storage driver, deployed a sample pod, and a sample PVC. Then, I created a VolumeSnapshotClass and a VolumeSnapshot to trigger snapshotting from the existing PVC. However, snapshotting didn't work. VolumeSnapshot's readyToUse field does not become true, and I don't see any related disk or snapshot on the GCP project.

Create a PVC and a Pod:

```yaml
kind: PersistentVolumeClaim
apiVersion: v1
metadata:
  name: source-pvc
spec:
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: 6Gi
---
apiVersion: v1
kind: Pod
metadata:
  name: source-pod
spec:
  containers:
  - name: busybox
    image: busybox
    command: ["/bin/sh", "-c"]
    args: ["touch /demo/data/sample-file.txt && sleep 3000"]
    volumeMounts:
    - name: source-data
      mountPath: /demo/data
  volumes:
  - name: source-data
    persistentVolumeClaim:
      claimName: source-pvc
      readOnly: false
```

Then, create a VolumeSnapshotClass and a VolumeSnapshot:

```yaml
apiVersion: snapshot.storage.k8s.io/v1beta1
kind: VolumeSnapshotClass
metadata:
  annotations:
    snapshot.storage.kubernetes.io/is-default-class: "true"
  name: default-snapshot-class
driver: pd.csi.storage.gke.io
deletionPolicy: Delete
---
apiVersion: snapshot.storage.k8s.io/v1beta1
kind: VolumeSnapshot
metadata:
  name: snapshot-source-pvc
spec:
  source:
    persistentVolumeClaimName: source-pvc
```

Check the VolumeSnapshot status:

`kubectl get volumesnapshot snapshot-source-pvc`

**Environment**:

- Gardener version (if relevant): v1.3.0
- Extension version:
- Kubernetes version (use `kubectl version`): 1.18.1
- Cloud provider or hardware configuration: GCP
- Others:

ialidzhikov:
Starging for v2.0.0 the external-snapshotter is split into 2 subcomponents - snapshot controller and CSI external-snapshotter sidecar. Ref https://github.com/kubernetes-csi/external-snapshotter/blob/master/CHANGELOG-2.0.md.
Currently we are missing and not deploying the snapshot controller which is responsible for the VolumeSnapshot watch.

Fixes #73

**Release note**:

```improvement operator
`gardener-extension-provider-gcp` now deploys required CSI snapshot controller and in this way fixes the issue with `VolumeSnapshot` resources which are not getting ready to use.
```


charts/internal/seed-controlplane/charts/csi-driver-controller/templates/csi-driver-controller.yaml
```diff
@@ -127,7 +127,7 @@ spec:
         - --kubeconfig=/var/lib/csi-snapshotter/kubeconfig
         - --leader-election
         - --leader-election-namespace=kube-system
-        - --snapshot-name-prefix={{ .Release.Namespace }}
+        - --snapshot-name-prefix=snapshot
```
rfranzke, line 130: Why is this changed? Is this backwards compatible?

ialidzhikov: > Why is this changed?

Because while testing I hit the same length constraint as the PVs - see https://github.com/gardener/gardener-extension-provider-gcp/pull/61#discussion_r407238133.

> Is this backwards compatible?
Generally not, but in our case it is as with the current master it is not possible to create `VolumeSnapshot`s at all.

rfranzke: Yes, right. Makes perfect sense. Thank you!

ialidzhikov:
> In our Slack conversations yesterday I have understood that the external-snapshotter should run in the shoot while the snapshot-controller runs in the seed, but I don't see this changed here.

Sorry, this was just a wrong guess from my side.

> So external-snapshotter stays in the csi-driver-controller deployment and we just need the additional snapshot-controller next to it?

Yep, the `external-snapshotter` sidecar stays in the csi-driver-controller Deployment as it needs to talks to the CSI endpoint.
`snapshot-controller` is responsible for managing the create and delete lifecycle of VolumeSnapshots. It does not need to talk to the CSI endpoint so that's why I added it as separate Deployment.

PR #79, opened 2020-04-24
Relax GCP validation to allow updates

### LINKED ISSUE #74: cloudNAT settings can't be changed
**What happened**:
I tried changing the cloudNAT setting in one of my teams cluster's shoot, as it is described here https://github.com/gardener/gardener-extension-provider-gcp/blob/master/docs/usage-as-end-user.md. When I tried to 'Save' my changes the following error appeared "admission webhook "validation.gcp.provider.extensions.gardener.cloud" denied the request: spec.provider.infrastructureConfig.networks: Invalid value: gcp.NetworkConfig{VPC:(*gcp.VPC)(nil), CloudNAT:(*gcp.CloudNAT)(0xc000499b20), Internal:(*string)(nil), Worker:"10.250.0.0/19", Workers:"", FlowLogs:(*gcp.FlowLogs)(nil)}: field is immutable"

I expect the modification of the cloudNAT settings to be allowed.

**Environment**:
- Extension version: 1.5.1
- Kubernetes version (use `kubectl version`): 1.16.8
- Others:

zanetworker:
This  PR provides explicit validation for GCP InfrastructureConfig updates, this would allow other non-validated fields to be updated. 

Fixes #74 

**Release note**:

```improvement operator
GCP validation now allows updates to fields such as CloudNAT and FlowLogs. 
```

### DIFF
--- a/pkg/apis/gcp/validation/infrastructure.go
+++ b/pkg/apis/gcp/validation/infrastructure.go
@@ -17,9 +17,10 @@ package validation
 import (
 	"reflect"
 
+	apivalidation "k8s.io/apimachinery/pkg/api/validation"
+
 	apisgcp "github.com/gardener/gardener-extension-provider-gcp/pkg/apis/gcp"
 	cidrvalidation "github.com/gardener/gardener/pkg/utils/validation/cidr"
-	apivalidation "k8s.io/apimachinery/pkg/api/validation"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/apis/gcp/validation/infrastructure.go
```diff
@@ -122,7 +123,16 @@ func ValidateInfrastructureConfig(infra *apisgcp.InfrastructureConfig, nodesCIDR
...
+
+	vpcPath := field.NewPath("networks.vpc")
+
+	oldVPC := oldConfig.Networks.VPC
+	newVPC := newConfig.Networks.VPC
+
+	if oldVPC != nil && newVPC != nil {
+		allErrs = append(allErrs, apivalidation.ValidateImmutableField(newVPC.Name, oldVPC.Name, vpcPath.Child("name"))...)
+		allErrs = append(allErrs, apivalidation.ValidateImmutableField(newVPC.CloudRouter, oldVPC.CloudRouter, vpcPath.Child("cloudRouter"))...)
+	}
```
rfranzke, line 135: We also should prevent changing [those fields](https://github.com/gardener/gardener-extension-provider-gcp/blob/master/pkg/apis/gcp/v1alpha1/types_infrastructure.go#L40-L47), or?

pkg/apis/gcp/validation/infrastructure.go
```diff
@@ -121,8 +122,21 @@ func ValidateInfrastructureConfig(infra *apisgcp.InfrastructureConfig, nodesCIDR
 
 // ValidateInfrastructureConfigUpdate validates a InfrastructureConfig object.
 func ValidateInfrastructureConfigUpdate(oldConfig, newConfig *apisgcp.InfrastructureConfig, fldPath *field.Path) field.ErrorList {
-	allErrs := field.ErrorList{}
-	allErrs = append(allErrs, apivalidation.ValidateImmutableField(newConfig.Networks, oldConfig.Networks, fldPath.Child("networks"))...)
+	var (
+		allErrs  = field.ErrorList{}
+		networks = fldPath.Child("networks")
```
rfranzke, line 127:
```suggestion
		networksPath = fldPath.Child("networks")
```

pkg/apis/gcp/validation/infrastructure.go
```diff
@@ -121,8 +122,21 @@ func ValidateInfrastructureConfig(infra *apisgcp.InfrastructureConfig, nodesCIDR
...
+		vpcPath  = networks.Child("vpc")
+	)
+
+	oldVPC := oldConfig.Networks.VPC
+	newVPC := newConfig.Networks.VPC
+
+	if oldVPC != nil && newVPC != nil {
```
rfranzke, line 134: Hm, this would allow to completely reset the VPC, i.e., to set `vpc=nil`.

pkg/apis/gcp/validation/infrastructure.go
rfranzke: What about the internal CIDR? Shouldn't it be immutable as well?

pkg/apis/gcp/validation/infrastructure_test.go
```diff
@@ -293,16 +301,59 @@ var _ = Describe("InfrastructureConfig validation", func() {
 			Expect(ValidateInfrastructureConfigUpdate(infrastructureConfig, infrastructureConfig, fldPath)).To(BeEmpty())
 		})
 
-		It("should forbid changing the network section", func() {
+		It("should allow changing cloudNAT details", func() {
```
rfranzke, line 304: Can you also add a test case that checks if changes to the flow logs are allowed? Or, do it like with the negative test, change this one here to try changing all the fields that should be allowed to change

PR #90, opened 2020-05-06
Support add custom label on GCP compute engine vm

### LINKED ISSUE #82: Support add custom label on GCP compute engine vm
In pkg/controller/worker/machines.go,  the machineClassSpec  labels map only contains one key named "name",  can we all pool labels to the machineClassSpec?  
```
	machineClassSpec := map[string]interface{}{
				"region":             w.worker.Spec.Region,
				"zone":               zone,
				"canIpForward":       true,
				"deletionProtection": false,
				"description":        fmt.Sprintf("Machine of Shoot %s created by machine-controller-manager.", w.worker.Name),
				"disks":              disks,
				"labels": map[string]interface{}{
					"name": w.worker.Name,
				},
```
By the way, AWS extension needs this feature too.

**Why is this needed**:
We need some custom labels to distinguish  VM, then we can charge our tenant by query the cost from the big query.

If it's possible, I can submit a PR to do this

sdojjy:
This PR will add worker pool labels to GCP VM, with this feature we can add custom labels to GCP VM

Fixes #82

**Release note**:

```improvement operator
WorkerPool labels will be added to GCP VM as the resource label
```

### DIFF
--- a/pkg/controller/worker/machines.go
+++ b/pkg/controller/worker/machines.go
@@ -18,6 +18,8 @@ import (
 	"context"
 	"fmt"
 	"path/filepath"
+	"regexp"
+	"strings"
 
 	"github.com/gardener/gardener/pkg/apis/extensions/v1alpha1"
 
@@ -34,6 +36,12 @@ import (
 	machinev1alpha1 "github.com/gardener/machine-controller-manager/pkg/apis/machine/v1alpha1"
 	"k8s.io/apimachinery/pkg/runtime"
 )
+
+var (
+	labelRegex = regexp.MustCompile(`[^a-z0-9_-]`)
+)
+
+const maxGcpLabelCharactersSize = 63
 
 // MachineClassKind yields the name of the GCP machine class.
 func (w *workerDelegate) MachineClassKind() string {
@@ -147,6 +155,7 @@ func (w *workerDelegate) generateMachineConfig(ctx context.Context) error {
 			disks = append(disks, disk)
 		}
 
+		gceInstanceLabels := getGceInstanceLabels(w.worker.Name, pool)
 		for zoneIndex, zone := range pool.Zones {
 			zoneIdx := int32(zoneIndex)
 			machineClassSpec := map[string]interface{}{
@@ -156,10 +165,8 @@ func (w *workerDelegate) generateMachineConfig(ctx context.Context) error {
 				"deletionProtection": false,
 				"description":        fmt.Sprintf("Machine of Shoot %s created by machine-controller-manager.", w.worker.Name),
 				"disks":              disks,
-				"labels": map[string]interface{}{
-					"name": w.worker.Name,
-				},
-				"machineType": pool.MachineType,
+				"labels":             gceInstanceLabels,
+				"machineType":        pool.MachineType,
 				"networkInterfaces": []map[string]interface{}{
 					{
 						"subnetwork":        nodesSubnet.Name,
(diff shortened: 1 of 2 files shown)


pkg/controller/worker/machines.go
```diff
@@ -147,6 +147,13 @@ func (w *workerDelegate) generateMachineConfig(ctx context.Context) error {
 			disks = append(disks, disk)
 		}
 
+		// gcp vm labels
+		machineClassLabels := map[string]interface{}{
```
rfranzke, line 151:
```suggestion
		gceInstanceLabels := map[string]interface{}{
```

pkg/controller/worker/machines.go
```diff
@@ -147,6 +147,13 @@ func (w *workerDelegate) generateMachineConfig(ctx context.Context) error {
 			disks = append(disks, disk)
 		}
 
+		// gcp vm labels
+		gceInstanceLabels := map[string]interface{}{
+			"name": w.worker.Name,
+		}
+		for k, v := range pool.Labels {
+			machineClassLabels[k] = v
```
rfranzke, line 155:
```suggestion
			gceInstanceLabels[k] = v
```

pkg/controller/worker/machines.go
```diff
@@ -156,10 +163,8 @@ func (w *workerDelegate) generateMachineConfig(ctx context.Context) error {
 				"deletionProtection": false,
 				"description":        fmt.Sprintf("Machine of Shoot %s created by machine-controller-manager.", w.worker.Name),
 				"disks":              disks,
-				"labels": map[string]interface{}{
-					"name": w.worker.Name,
-				},
-				"machineType": pool.MachineType,
+				"labels":             machineClassLabels,
```
rfranzke, line 166:
```suggestion
				"labels":             gceInstanceLabels,
```

sdojjy:
> Can you adapt the unit test accordingly?

OK

rfranzke:
Looks good, thanks! Can you squash the commits into a single one so that we can go ahead with the merge? :)

rfranzke:
Thanks! Looks good, need to play with it - will come back

sdojjy:
> I tried it out, but I ran into a problem. When using these labels
> 
> ```yaml
>   labels:
>     bar: baz-foo_bar.foobar
>     foo: bar
> ```
> 
> then the GCE API rejects the VM creation:
> 
> ```
> E0507 04:43:37.460324       1 machine.go:408] Error while creating machine shoot--proj--shootcpu-worker-z1-7bdb7b8768-qh74k: googleapi: Error 400: Invalid value for field 'resource.labels': ''. Label value 'baz-foo_bar.foobar' violates format constraints. The value can only contain lowercase letters, numeric characters, underscores and dashes. The value can be at most 63 characters long. International characters are allowed., invalid
> ```
> 
> I think you need to sanitize the labels before putting them into the machine class.

thanks, I will update this PR

pkg/controller/worker/machines.go
```diff
@@ -247,3 +260,12 @@ func createDiskSpec(volume v1alpha1.Volume, workerName string, machineImage stri
 
 	return disk, nil
 }
+
+// SanitizeGcpLabel will sanitize the label/value base on the gcp label Restrictions
+func SanitizeGcpLabel(label string) string {
```
rfranzke, line 265: Thank you! Looks good, just a unit test for this function is missing. :)

sdojjy: good suggestion. done

rfranzke:
@sdojjy I think you have to rebase, sorry, could you please do so that CI can pass?

sdojjy:
> @sdojjy I think you have to rebase, sorry, could you please do so that CI can pass?

OK

sdojjy:
CI will take more than 3 hours?

PR #112, opened 2020-06-09
Add validation for cloud provider Secrets

ialidzhikov:

/area robustness
/area ops-productivity
/kind enhancement
/priority normal
/platform gcp

Add validation for cloud provider Secrets.

Ref https://github.com/gardener/gardener/issues/2293

**Release note**:

```improvement operator
The existing ValidatingWebhookConfiguration of `admission-gcp` for Shoot validation does now validate also the Shoot secret. `admission-gcp` does now feature also a new webhook that prevents Shoot secret to be updated with invalid keys.
```


ialidzhikov:
Something interesting that I figure out today trying to use `client.MatchingFields/client.MatchingFieldsSelector` with the controller-runtime client.
When the client is a non-cached one, the above options are handled as field selector sent to the API server.
When the client is a cached one, the above options work on index and you have to explicitly add the index fields. For example:
```golang
			if err := mgr.GetFieldIndexer().IndexField(&gardencorev1beta1.Shoot{}, "spec.seedName", func(rawObj runtime.Object) []string {
				shoot := rawObj.(*gardencorev1beta1.Shoot)
				if shoot.Spec.SeedName == nil {
					return nil
				}

				return []string{*shoot.Spec.SeedName}
			}); err != nil {
				return err
			}
```

Otherwise the calls always fail with 

```
Index with name field:spec.seedName does not exist
```

Ahh, that was quite unintuitive for me, as I added the required changes for gardener-apiserver for new field-selectors, kubectl and a non-cached client were working well, but I was wondering why the cached client continues to return `index does not exist`.

pkg/admission/validator/secret.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	return nil
+}
+
+func isSecretUsedByShoot(ctx context.Context, c client.Client, secret *corev1.Secret, providerType string) (bool, error) {
+	// controller-runtime cached client does not support non-exact field matches
+	secretBindings := &gardencorev1beta1.SecretBindingList{}
+	if err := c.List(ctx, secretBindings,
+		client.MatchingFields{"secretRef.namespace": secret.Namespace}); err != nil {
```
vpnachev, line 71: you could use the constant `SecretRefNamespaceField` from pkg/admission/cache

pkg/admission/validator/secret.go
```diff
@@ -0,0 +1,94 @@
...
+		return false, err
+	}
+
+	for _, secretBinding := range secretBindings.Items {
+		// Filter out the the SecretBindings that do not reference the given secret
+		if secretBinding.SecretRef.Name != secret.Name {
```
vpnachev, line 77: I think you can also compare the namespace names, am I right?
```suggestion
		if secretBinding.SecretRef.Name != secret.Name || secretBinding.SecretRef.Namespace != secret.Namespace {
```

pkg/admission/validator/secret.go
```diff
@@ -0,0 +1,94 @@
...
+			continue
+		}
+
+		shoots := &gardencorev1beta1.ShootList{}
+		if err := c.List(ctx, shoots, client.MatchingFields{"spec.secretBindingName": secretBinding.Name}); err != nil {
```
vpnachev, line 82: the constant `SecretBindingNameField` from pkg/admission/cache can be used here.

pkg/admission/validator/secret.go
```diff
@@ -0,0 +1,94 @@
...
+			return false, err
+		}
+
+		for _, shoot := range shoots.Items {
+			if shoot.Spec.Provider.Type == providerType {
```
vpnachev, line 87: I think you need to check that shoot.Namespace and secretBinding.Namespace are equal.
This is needed to catch the below case.

| Project     | Shoot       | SecretBinding  | CloudProvider |
|:------------- |:-------------|:-----|:------------------------|
| foo1      | bar | baz | aws |
| foo2      | bar      |   baz | gcp |

example/40-validatingwebhookconfiguration.yaml
```diff
@@ -22,4 +22,24 @@ webhooks:
   sideEffects: None
   clientConfig:
     url: "https://localhost:9443/webhooks/validate"
```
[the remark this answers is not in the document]
ialidzhikov: There is no need to adapt anything. The gardener-apiserver is running on localhost and it can reach the webhook server running also on localhost. The resource for the newly introduced webhook is Secret and the kube-apiserver (if minikube is used) cannot by default reach the webhook on localhost. That's why `hack/dev-setup-admission-gcp.sh` is introduced.

pkg/admission/validator/secret.go
ialidzhikov: These secretBindings are already fetched with `client.MatchingFields{"secretRef.namespace": secret.Namespace}`. There is no need to check one more time that the secret namespace is matching the secretBinging secretRef.namespace. (Actually this if is needed because the cached client does support only 1 field selector :( )

pkg/admission/cache/informer_cache.go
```diff
@@ -0,0 +1,42 @@
...
+
+import (
+	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+// SecretRefNamespaceField is the field name for the index function that extracts the corresponding field from SecretBinding.
+const SecretRefNamespaceField string = "secretRef.namespace"
+
+func SecretRefNamespaceIndexerFunc(rawObj runtime.Object) []string {
```
rfranzke, line 25: doc string

pkg/admission/cache/informer_cache.go
```diff
@@ -0,0 +1,42 @@
...
+	if !ok {
+		return []string{}
+	}
+	return []string{secretBinding.SecretRef.Namespace}
+}
+
+// SecretBindingNameField is the field name for the index function that extracts the corresponding field from Shoot.
+const SecretBindingNameField string = "spec.secretBindingName"
+
+func SecretBindingNameIndexerFunc(rawObj runtime.Object) []string {
```
rfranzke, line 36: doc string

pkg/apis/gcp/validation/secret.go
```diff
@@ -0,0 +1,44 @@
...
+	"regexp"
+
+	"github.com/gardener/gardener-extension-provider-gcp/pkg/gcp"
+	"github.com/gardener/gardener-extension-provider-gcp/pkg/internal"
+	corev1 "k8s.io/api/core/v1"
+)
+
+var projectIDRegexp = regexp.MustCompile(`^(?P<project>[a-z][a-z0-9-]{4,28}[a-z0-9])$`)
+
+func ValidateCloudProviderSecret(secret *corev1.Secret) error {
```
rfranzke, line 28: doc string

example/40-validatingwebhookconfiguration.yaml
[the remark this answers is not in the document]
timuthy: Hm, doesn't it make sense to also point to the `gardener-extension-admission-gcp` service for consistency reasons?

pkg/admission/validator/secret.go
timuthy: > (Actually this if is needed because the cached client does support only 1 field selector :( )

Does it make sense to add a TODO and come back later when the controller-runtime supports multiple?

pkg/admission/validator/secret.go
timuthy: > I think you need to check that shoot.Namespace and secretBinding.Namespace are equal.

In this case, isn't it only important to know if there is at least one shoot of type `gcp` which uses this secret? I don't see the missing use case, atm.

pkg/admission/validator/shoot.go
```diff
@@ -222,3 +235,36 @@ func newValidationContext(ctx context.Context, decoder runtime.Decoder, c client
...
+		secretKey = kutil.Key(secretBinding.SecretRef.Namespace, secretBinding.SecretRef.Name)
+	)
+	if err := lookup(ctx, s.client, s.apiReader, secretKey, secret); client.IgnoreNotFound(err) != nil {
+		return err
+	}
+
+	return gcpvalidation.ValidateCloudProviderSecret(secret)
+}
+
+func lookup(ctx context.Context, c client.Client, apiReader client.Reader, key client.ObjectKey, obj runtime.Object) error {
```
timuthy, line 259: Nice function which could be useful in the g/g extensions lib.

pkg/admission/validator/webhook.go
```diff
@@ -43,3 +46,17 @@ func New(mgr manager.Manager) (*extensionswebhook.Webhook, error) {
 		},
 	})
 }
+
+// NewSecretsWebhook creates a new validation webhook for Secrets.
+func NewSecretsWebhook(mgr manager.Manager) (*extensionswebhook.Webhook, error) {
```
timuthy, line 51: Was it necessary to create a new endpoint or why didn't you just add the `NewSecretValidator()` to the existing one?

pkg/admission/validator/secret.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+// InjectClient injects the given client into the validator.
+func (s *secret) InjectClient(client client.Client) error {
+	s.client = client
+	return nil
+}
+
+// Validate validates the given secret objects.
+func (s *secret) Validate(ctx context.Context, new, _ runtime.Object) error {
```
timuthy, line 47: Don't you want to compare the old and new secret and skip any further logic in case both are equal?

pkg/admission/validator/secret.go
```diff
@@ -0,0 +1,94 @@
...
+	if isUsed {
+		if err := gcpvalidation.ValidateCloudProviderSecret(secret); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func isSecretUsedByShoot(ctx context.Context, c client.Client, secret *corev1.Secret, providerType string) (bool, error) {
```
timuthy, line 67: Can you add unit tests for this function?

pkg/admission/validator/secret.go
vpnachev: > These secretBindings are already fetched with client.MatchingFields{"secretRef.namespace": secret.Namespace}. 

You are right, I failed to see it :smile:

pkg/admission/validator/webhook.go
ialidzhikov: Yep, generally it would be good to have only 1 endpoint and I recall I had difficulties naming the new one. The issue currently is that the existing webhook handler specifies 

```
		Predicates: []predicate.Predicate{extensionspredicate.GardenCoreProviderType(gcp.Type)},
```

which is perfectly fine for the Shoot and CloudProfile but it will return always `false` for `corev1.Secret`.

timuthy: Oh, I see. Thanks for explaining!

pkg/admission/cache/informer_cache.go
ialidzhikov, line 26: These funcs are not admission-gcp specific, they would also need to go under g/g extensions pkgs to make them usable also from other provider extensions.

pkg/admission/validator/secret.go
ialidzhikov: Yep. This func would also need to go under g/g extension pkg to make it usable also from other provider extensions.

docs/local-setup.md
```diff
@@ -0,0 +1,27 @@
...
+3. (Optional) Setup the webhook service.
+
+```bash
+./hack/dev-setup-admission-gcp.sh
+```
+
+4. Apply the `ValidatingWebhookConfiguration`
+
+```bash
+kubectl apply -f example/40-validatingwebhookconfiguration.yaml
```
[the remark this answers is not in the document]
ialidzhikov: I moved it to `hack/dev-setup-admission-gcp.sh`.

example/40-validatingwebhookconfiguration.yaml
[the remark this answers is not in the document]
ialidzhikov: I have no strong opinion, I now modified both to point to the service.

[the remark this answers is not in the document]
ialidzhikov: Actually it wouldn't work with the service approach for local setup which is using minikube for example as the gardener-apiserver is not able to call the service. Probably something similar would work with the nodeless setup?
I am bringing it back to url for now.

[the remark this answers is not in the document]
timuthy: Yes makes sense, thank you.

docs/local-setup.md
```diff
@@ -0,0 +1,27 @@
...
+
+    Make sure that the `KUBECONFIG` environment variable is pointing to the local garden cluster.
+
+    ```bash
+    make start-admission
+    ```
+
+1. Setup the `ValidatingWebhookConfiguration`.
+
+    `hack/dev-setup-admission-gcp.sh` will configure the webhook Service which will allow to the kube-apiserver of your local cluster to reach the webhook server. It will also apply the `ValidatingWebhookConfiguration` manifest.
```
timuthy, line 21:
```suggestion
    `hack/dev-setup-admission-gcp.sh` will configure the webhook Service which will allow the kube-apiserver of your local cluster to reach the webhook server. It will also apply the `ValidatingWebhookConfiguration` manifest.
```

ialidzhikov:
> Nice. Can you update the g/g dependency to include [gardener/gardener#2537](https://github.com/gardener/gardener/pull/2537) and adapt this PR to it?

Sorry for the long delay, now I updated the PR.

PR #117, opened 2020-06-18
Remove IPIP tunnels from GCP Calico Networking

### LINKED ISSUE #115: Remove IPIP Tunnels from GCP
/area networking
/kind enhancement

Today, the `cloud-controller-manager` is used to configure routes on GCP (`configure-cloud-routes=true`), however these routes are not used since we are utilizing CNI tunnels (e.g., IPIP or VxLAN) to enabled pod-connectivity cross nodes and zones. 

To completely remove tunnels for GCP, we would need to disable IPIP tunnels on Calico (e.g., via a webhook similar to Azure Setup `backend=none`). Furthermore, we need to deploy [`ip-masq-agent`](https://github.com/kubernetes-sigs/ip-masq-agent) to prevent Pod generated traffic from being blocked by the GCP infrastructure . 

**Why is this needed**:
- Improve pod routing performance. 
- Avoid possible overlaps between tunnel and cluster CIDRs.
- Eliminate MTU inconsistencies.

zanetworker:

/area networking
/kind enhancement
/priority normal
/platform gcp

This PR removes tunnel usage in GCP by setting Calico backend to `none` and adding an `ip-masq-agent` that NATs traffic to the outside. More information in #115. 

Fixes #115 

**Release note**:

```improvement operator
GCP networking with Calico is now tunnel-free (i.e., no IPIP cross communication between nodes). For more information, please check the documentation (docs/usage-as-end-user.md).
```


go.mod
```diff
@@ -9,6 +9,7 @@ require (
 	github.com/frankban/quicktest v1.9.0 // indirect
 	github.com/gardener/etcd-druid v0.3.0
 	github.com/gardener/gardener v1.6.0
+	github.com/gardener/gardener-extension-networking-calico v1.8.1-0.20200618083450-4112b60d1bac
```
rfranzke, line 12: Can you reference a released version?

pkg/webhook/network/mutate.go
```diff
@@ -0,0 +1,56 @@
+// Copyright (c) 2019 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
rfranzke, line 1:
```suggestion
// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```

charts/images.yaml
```diff
@@ -55,3 +55,9 @@ images:
   sourceRepository: github.com/kubernetes-csi/livenessprobe
   repository: quay.io/k8scsi/livenessprobe
   tag: "v2.0.0"
+
+
+- name: ip-masq-agent
+  sourceRepository: github.com/kubernetes-sigs/ip-masq-agent
+  repository:  k8s.gcr.io/ip-masq-agent-amd64
```
rfranzke, line 62:
```suggestion
  repository: k8s.gcr.io/ip-masq-agent-amd64
```

charts/images.yaml
```diff
@@ -55,3 +55,9 @@ images:
   sourceRepository: github.com/kubernetes-csi/livenessprobe
   repository: quay.io/k8scsi/livenessprobe
   tag: "v2.0.0"
+
+
```
rfranzke, line 59:
```suggestion
```

charts/internal/shoot-system-components/charts/ip-masq-agent/templates/daemonset.yaml
```diff
@@ -0,0 +1,50 @@
...
+            privileged: true
+{{- if .Values.resources }}
+          resources:
+{{ toYaml .Values.resources | indent 12 }}
+{{- end }}
+          volumeMounts:
+            - mountPath: /etc/config
+              name: config
+      volumes:
+        - configMap:
```
rfranzke, line 43: Where is this config coming from? I don't see it being created anywhere.

go.mod
zanetworker: no released version with this, figured this dependency after I did the release (gardener is vendored to 1.6 here which brought the generic actuator refactor in). Or I do another major release with 1 commit ...

charts/internal/shoot-system-components/charts/ip-masq-agent/templates/daemonset.yaml
zanetworker: its not, its optional later if live-config is required, have a look [here](https://github.com/kubernetes-sigs/ip-masq-agent) "As a result, the agent can be reconfigured in a live cluster by creating or editing this ConfigMap"

zanetworker:
> Looks good overall. Is this something that end-users should be aware of (as it's not configurable and rolled out to all shoots by default)? Does it make sense to mention it in the `docs/usage-as-end-user.md`?

the change should be transparent to the users, so there is no action required but probably a good to know. I guess the release note should be enough as a disclaimer?

go.mod
rfranzke: OK, we should update it then when a new version is available. It should not be critical as only APIs are used. OK for now then.

charts/internal/shoot-system-components/charts/ip-masq-agent/templates/daemonset.yaml
rfranzke: OK, interesting. I think it's not obvious from the manifest itself, so probably a comment would help to understand what is going on with this configmap.

rfranzke:
> the change should be transparent to the users, so there is no action required but probably a good to know. I guess the release note should be enough as a disclaimer?

Great, yes, should be enough then. Does it make sense to make people aware that they can (re)configure this `ip-masq-agent` with this config map?

zanetworker:
@rfranzke My last push was to rebase, as g/g was  vendored to 1.6.3 instead.

PR #175, opened 2020-09-15
Migration from in-tree to out-of-tree MCM controllers

### LINKED ISSUE #174: Move from in-tree to out-of-tree MCM

/area dev-productivity
/kind enhancement
/priority normal
/platform gcp

We would like to split the MCM code base into two different pods using the OOT approach in MCM - https://github.com/gardener/machine-controller-manager/pull/460. 

**Why is this needed**:
This would ease the maintenance of the provider specific MCM code independently.

prashanth26:

/area dev-productivity
/kind enhancement
/priority normal
/platform gcp

We would like to split the MCM code base into two different pods using the OOT approach in MCM - https://github.com/gardener/machine-controller-manager/pull/460.

Fixes #174

The following is the approach that I have used,
- Find any existing old machineClass CRDs (GCPMachineClasses), and set deletion timestamps on them during deploy machineClass.
- Update the charts to deploy the new machine class instead of the older one. 

I will update the MCM-GCP provider image by the time this review is done. Also, add and update the tests accordingly. In the meantime, opened up this PR for your opinion on the approach and to identify if anything could be done better. 

**Release note**:

```noteworthy operator
The worker controller now deploys MachineClass CR instead of GCPMachineClass. It also sets the deletion timestamp on any existing old GCPMachineClass CRs. Any migration of existing GCPMachineClass to the newMachineClass occurs implicitly without causing rollouts of existing nodes/VMs.
```
```noteworthy operator
The MCM is deployed using the out-of-tree style of deployment. More details can be at [MCM PR#460](https://github.com/gardener/machine-controller-manager/pull/460).
```

### DIFF
--- a/pkg/controller/worker/machines.go
+++ b/pkg/controller/worker/machines.go
@@ -22,6 +22,8 @@ import (
 	"strings"
 
 	"github.com/gardener/gardener/pkg/apis/extensions/v1alpha1"
+	"github.com/pkg/errors"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	apisgcp "github.com/gardener/gardener-extension-provider-gcp/pkg/apis/gcp"
 	gcpapi "github.com/gardener/gardener-extension-provider-gcp/pkg/apis/gcp"
@@ -34,6 +36,7 @@ import (
 	"github.com/gardener/gardener/pkg/client/kubernetes"
 	machinev1alpha1 "github.com/gardener/machine-controller-manager/pkg/apis/machine/v1alpha1"
 	computev1 "google.golang.org/api/compute/v1"
+	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
@@ -43,14 +46,25 @@ var (
 
 const maxGcpLabelCharactersSize = 63
 
-// MachineClassKind yields the name of the GCP machine class.
+// MachineClassKind yields the name of the machine class kind used by GCP provider.
 func (w *workerDelegate) MachineClassKind() string {
-	return "GCPMachineClass"
+	return "MachineClass"
 }
 
-// MachineClassList yields a newly initialized GCPMachineClassList object.
+// MachineClassList yields a newly initialized MachineClassList object.
 func (w *workerDelegate) MachineClassList() runtime.Object {
-	return &machinev1alpha1.GCPMachineClassList{}
+	return &machinev1alpha1.MachineClassList{}
+}
+
+// cleanupOldMachineClasses sets deletionTimestamp on any older version machine class CRs.
+func (w *workerDelegate) cleanupOldMachineClasses(ctx context.Context, namespace string, machineClassList runtime.Object, wantedMachineDeployments worker.MachineDeployments) error {
+	if err := w.Client().List(ctx, machineClassList, client.InNamespace(namespace)); err != nil {
+		return err
+	}
+
+	return meta.EachListItem(machineClassList, func(machineClass runtime.Object) error {
+		return w.Client().Delete(ctx, machineClass)
+	})
 }
 
 // DeployMachineClasses generates and creates the GCP specific machine classes.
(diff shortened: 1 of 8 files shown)


prashanth26:
reviewed/do-not-merge

pkg/controller/worker/machines.go
```diff
@@ -43,14 +46,29 @@ var (
...
+	if err := w.Client().List(ctx, machineClassList, client.InNamespace(namespace)); err != nil {
+		return err
+	}
+
+	return meta.EachListItem(machineClassList, func(machineClass runtime.Object) error {
+		if err := w.Client().Delete(ctx, machineClass); err != nil {
+			return err
+		}
+
+		return nil
```
rfranzke, line 70:
```suggestion
		return w.Client().Delete(ctx, machineClass)
```

pkg/controller/worker/machines_test.go
```diff
@@ -494,7 +494,7 @@ var _ = Describe("Machines", func() {
 						},
 					}
 				}
-				It("should return the expected machine deployments when disableExternal IP is true with profile image types", func() {
+				XIt("should return the expected machine deployments when disableExternal IP is true with profile image types", func() {
```
rfranzke, line 497: XIt? :)

rfranzke (commented):
Is there a document that explains how the migration works? :)

prashanth26 (commented):
There are two parts to migration 

Migration of GCPMachineClass to MachineClass only using MCM (without Gardener). This is something that has been documented - [here](https://github.com/gardener/machine-controller-manager/blob/master/docs/development/machine_error_codes.md#generatemachineclassformigration) in the MCM documentation. 

However, in the context of Gardener, the migration is done by Gardener and not by MCM. I haven't documented anything until now, But this something how the flow looks like from Gardener point of view (I can document this)

- The MCM deployment is first updated by the Gardener to deploy 2 pods, MCM and Machine-controller-provider and will wait for the rolling update to go through - https://github.com/gardener/gardener/blob/master/extensions/pkg/controller/worker/genericactuator/machine_controller_manager.go#L84-L86. 
- The worker controller first sets deletion timestamps on existing GCPMachineClasses. The MCM (in tree GCPMachineClass reconciler) in turn will wait for all references to these GCPMachineClasses to change to zero after which it deletes them.
- The extension chart then deploys the new `MachineClass` objects corresponding to the `charts`. And MCM (MachineDeployment) code is modified to avoid rolling updates on changing of just classKind like here - https://github.com/gardener/gardener/blob/master/extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go#L298.

pkg/controller/worker/machines_test.go
prashanth26: Yes, there was an issue while mocking the ListGCPMachinesCall. I have fixed it and now all the existing tests are working and this line is no longer `XIt()` :)

rfranzke:
Thanks @prashanth26. I guess this will not result in a rolling update of the machines, or?

prashanth26:
> Thanks @prashanth26. I guess this will not result in a rolling update of the machines, or?

Yes, you are right. This will not result in a rolling update. Existing machines will continue to run as it is today. The MCM logic is modified to exactly act accordingly.

hardikdr:
Btw how does it affect the clusters with ongoing roll-outs ? Probably related more to the mcm-gcp-provider, but just curious.

Minor nit in releae note: `s/set's/sets/g`

prashanth26:
> Minor nit in releae note: `s/set's/sets/g`

Updated the release note. 

> Btw how does it affect the clusters with ongoing roll-outs? Probably related more to the mcm-gcp-provider, but just curious.

Ongoing cluster rollouts should continue the rollout as it is today. As we are only updating the `kind` field that no longer causes rollouts. However, I shall verify the same by running local tests as well.

hardikdr (approved):
/lgtm

rfranzke (approved):
/lgtm

prashanth26:
/unhold i have updated the docker images to release versions. I have updated the release notes to keep it more meaningful. 

I was wondering if I should create any explicit doc to explain the process of migration in the gardener-extension-gcp context? if yes, should I add it as a new file under the docs/ section?

rfranzke:
> I was wondering if I should create any explicit doc to explain the process of migration in the gardener-extension-gcp context? if yes, should I add it as a new file under the docs/ section? or is the release notes understandable and enough?

Hm, there is not much GCP-specific "process", right? IMO no additional doc is necessary. You can have a pointer to a general migration document which is maintained in the MCM repo in this release note here.

PR #217, opened 2020-12-01
Readd role label key to pod template of mcm and ccm

ialidzhikov:
/kind bug

In #214 I missed that the dependency-watchdog-endpoint component is currently using `garden.sapcloud.io/role` to select pods to restart in case the kube-apiserver is in CrashLoopBackOff and it is recovered after it. Let's readd the label to prevent regressing this behaviour of dependency-watchdog-endpoint. 
I added the label only to the pod template and not to the deployment to prevent shoot care controller of gardenlet to select the deployments for healthchecks (their healthchecks are already performed by the extension itself).

**Release note**:

```other operator
NONE
```


DockToFuture (approved):
/lgtm

PR #222, opened 2020-12-18
update mcm and use worker secret ref. for the machine class credentialsSecretRef

danielfoehrKn:

/area quality
/kind enhancement
/priority normal
/platform gcp

Uses the machine classes  .{spec.}credentialsSecretRef field. It is a reference to a secret containing only the credentials, while today's .{spec.}secretRef still contains the user-data. See [here for more details](https://github.com/gardener/machine-controller-manager/pull/578).

Uses the worker's secret reference as the credentialsSecretRef.
This means all worker pools use the same "cloudprovider" secret containing only the cloud provider credentials.
The existing MachineClass SecretReference then only contains the user data that is different for each pool. 

Fixes #

**Release note**:

```other operator
Reducing credential update complexity by all the machine classes using the new .{spec.}credentialsSecretRef field.
This means all worker pools use the same "cloudprovider" secret containing only the cloud provider credentials.
The existing MachineClass SecretReference only contains the user data that is different for each pool.
```


rfranzke (approved):
/lgtm
/invite @prashanth26 
/assign @prashanth26

prashanth26 (commented):
As the master is using the MCM-provider-gcp with the required changes here - https://github.com/gardener/gardener-extension-provider-gcp/blob/master/charts/images.yaml#L38 & https://github.com/gardener/machine-controller-manager-provider-gcp/commit/5783306d4bdc0b9c2318a84193b020dd19ad60f9. Looks good to me.

prashanth26 (approved):

PR #234, opened 2021-02-05
VPA for validator/admission component

rfranzke:

/area scalability auto-scaling
/kind enhancement
/priority normal
/platform gcp

VPA for validator/admission component

**Release note**:

```noteworthy operator
The validator/admission component's Helm chart is now deploying a `VerticalPodAutoscaler` resource by default. If undesired or no VPA is available in the garden cluster then it can be turned of via `.Values.global.vpa.enabled=false`.
```

### DIFF
--- a/charts/gardener-extension-admission-gcp/values.yaml
+++ b/charts/gardener-extension-admission-gcp/values.yaml
@@ -7,6 +7,14 @@ global:
     pullPolicy: IfNotPresent
   replicaCount: 1
   resources: {}
+  vpa:
+    enabled: true
+    resourcePolicy:
+      minAllowed:
+        cpu: 50m
+        memory: 64Mi
+    updatePolicy:
+      updateMode: "Auto"
   webhookConfig:
     caBundle: |
       -----BEGIN CERTIFICATE-----
--- a/cmd/gardener-extension-admission-gcp/app/app.go
+++ b/cmd/gardener-extension-admission-gcp/app/app.go
@@ -29,6 +29,7 @@ import (
 	"github.com/gardener/gardener/pkg/apis/core/install"
 	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
 	"github.com/spf13/cobra"
+	corev1 "k8s.io/api/core/v1"
 	componentbaseconfig "k8s.io/component-base/config"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
@@ -91,6 +92,16 @@ func NewAdmissionCommand(ctx context.Context) *cobra.Command {
 				return err
 			}
 
+			// List all secrets to implicitly start a WATCH. This will fill up the webhook server's caches (hence, increasing
+			// its memory usage) right at the start-up. This is done to prevent unpredictable load spikes when the server receives
+			// its first request (without this list it would only then start the WATCH, leading to a huge memory increase on large
+			// landscapes. Hence, this "early listing" should prevent the server from being OOMKilled and (potentially multiple times)
+			// scaled by VPA (we have observed too "optimistic" up-scaling behavior of VPA in such cases).
+			// See https://github.com/gardener/gardener-extension-provider-gcp/pull/234 for more details.
+			if err := mgr.Add(&secretLister{mgr}); err != nil {
+				return err
+			}
+
 			return mgr.Start(ctx)
 		},
 	}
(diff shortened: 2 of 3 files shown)


ialidzhikov:
What changed after https://github.com/gardener/gardener-extension-provider-gcp/pull/141?

rfranzke:
Well, you know best after you had to manually increase the resource limits yesterday in one of the landscapes, right 😉 

From https://github.com/gardener/gardener-extension-provider-gcp/pull/141#issuecomment-664415250:

> I'm mainly worried about downscaling actions performed by VPA. In operations, we observed cases for which a proper upscaling didn't happen after downscaling and in order to recover we needed to delete the VPA object.

Hm, true, OTOH, we use VPA for basically everything and run thousands of clusters, so not sure if this component now is so special. Also, this statement is somewhat older and VPA became better in the meantime. Anyways, the feature is configurable, so if some landscape operator does not like it or it causes too much trouble then it can still be disabled and the old behaviour is preserved. Generally, offering VPA (and also enabling it by default) makes sense.

> The admission component directly affects the availability of the Shoot API and thus I'm a bit critical.

True, however, it's more critical if the statically configured resource requirements are not enough anymore and someone gets paged during the night for an incident that could have also auto-resolved via VPA.

/cc @timuthy

timuthy:
Can we combine this change with a secret sync on webhook start? Then we'd at least avoid a resource spike when the webhooks gets the first request after a while.

rfranzke:
@timuthy what do you have in mind? Simply listing all secrets on start-up?

timuthy:
> @timuthy what do you have in mind? Simply listing all secrets on start-up?

Yes, I think this will do the trick.

rfranzke:
@timuthy Here you go: 2645570

/squash

prashanth26 (approved):
/lgtm

Thanks for this.

charts/gardener-extension-admission-gcp/values.yaml
```diff
@@ -7,6 +7,14 @@ global:
     pullPolicy: IfNotPresent
   replicaCount: 1
   resources: {}
+  vpa:
+    enabled: true
+  # resourcePolicy:
+  #   minAllowed:
+  #     cpu: 100m
+  #     memory: 128Mi
```
vpnachev, line 15: Let's have some minAllowed by default. I remember we had issues with pods scaled down so much that then the VPA was not able to scale-up them anymore.

rfranzke: Which values do you propose? I can add something here, but I'm not sure whether it's too helpful. It highly depends on the landscape's size, so any default here will only work for small/"hello world" landscapes, I believe.

vpnachev: I believe those are already good, they just need to be uncommented.

rfranzke: Hm, still not sure. This will unnecessarily increase the values on small landscapes. For example, this is the current memory usage for a landscape with 180 shoots, 22 seeds, and ~1k secrets.

```
gardener-extension-admission-gcp-7dd46c9ddf-fgfbl         1m           27Mi
gardener-extension-admission-gcp-7dd46c9ddf-kcqkm         1m           26Mi
gardener-extension-admission-gcp-7dd46c9ddf-p4697         6m           26Mi
```

We can have rather small "min allowed" values (`10m`/`30Mi`, for example), they should be fine for most landscapes. For larger landscapes, the values need to be overridden anyways.
WDYT?

vpnachev: This is running just once per landscape, so it will not save so much resources. While, the lower the values, the higher the risk the VPA to stuck and to not up-scale the resources. I would set it to at least 50m CPU and 64MiB of memory.

cmd/gardener-extension-admission-gcp/app/app.go
```diff
@@ -91,6 +92,16 @@ func NewAdmissionCommand(ctx context.Context) *cobra.Command {
 				return err
 			}
 
+			// List all secrets to implicitly start a WATCH. This will fill up the webhook server's caches (hence, increasing
+			// its memory usage) right at the start-up. This is done done to prevent unpredictable load spikes when the server receives
```
timuthy, line 96:
```suggestion
			// its memory usage) right at the start-up. This is done to prevent unpredictable load spikes when the server receives
```

vpnachev (approved):
/lgtm

PR #249, opened 2021-03-31
Deserialize custom resources with EnableStrict

stoyanr:

/area control-plane
/kind enhancement
/priority 3
/platform gcp

* Adds the `EnableStrict` option to all places where a codec factory is created, to ensure that extension resource configs are always deserialized in "strict" mode. This means that resources with fields that are not allowed by the API schema will be rejected by validation and will cause errors (categorized as `ERR_CONFIGURATION_PROBLEM`) during reconciliation as well.
* Vendors g/g based on https://github.com/gardener/gardener/pull/3804 to ensure that a strict decoder is used also for `ControlPlaneConfig` and `WorkerConfig`, and errors are correctly categorized.

Fixes https://github.com/gardener/gardener-extension-provider-azure/pull/270, see also https://github.com/gardener/gardener-extension-provider-azure/pull/271

* The CI build fails because the vendored g/g requires go 1.16. This PR has to be force-merged, and I will later update the `branch.cfg`.

**Release note**:

```breaking user
Extension resource configs (`InfrastructureConfig`, `ControlPlaneConfigs`, `WorkerConfig`) are now deserialized in "strict" mode, including during validation by the admission webhook. This means that resources with fields that are not allowed by the API schema will be rejected by validation. Creating new shoots containing such resources will not be possible, and reconciling existing shoots will fail with an appropriate error until you manually update the shoot to make sure any extension resource configs contained in it are valid.
```


stoyanr:
I tested this thoroughly after revendoring https://github.com/gardener/gardener/pull/3804. It seems to work as expected. 

* It's not possible to create a new shoot with unknown fields in extension resource configs (`InfrastructureConfig`, `ControlPlaneConfig`, `WorkerConfig`), or to update an existing one by adding such unknown fields. Such shoots are rejected by the admission webhook.
* It's still possible to update an existing shoot with such unknown fields with a shoot that doesn't have them. The old shoot is decoded using a "lenient" decoder to make sure this can succeed.
* For existing shoots, unknown fields in all extension resource configs (`InfrastructureConfig`, `ControlPlaneConfig`, `WorkerConfig`) cause errors during reconciliation, and all such errors are correctly categorized as `ERR_CONFIGURATION_PROBLEM`, so the reconciliation fails without retry. Removing the unknown fields and reapplying the shoot leads to successful reconciliation (perhaps after a retry).

prashanth26:
Probably need to run `make revendor` as verify step is failing.

stoyanr:
The CI build fails because the vendored g/g requires go 1.16. This PR has to be force-merged, and I will later update the `branch.cfg`.

prashanth26:
>The CI build fails because the vendored g/g requires go 1.16. This PR has to be force-merged, and I will later update the branch.cfg.

Yes sorry, my bad. Realized it later.

Issue #143, opened 2020-07-28
Reduce memory footprint for GCP admission component


timuthy:
**How to categorize this issue?**

/area robustness
/area cost
/priority critical
/platform gcp

**What would you like to be added**:
Since the validation of cloud provider secrets has been introduced (#112) the GCP admission plugin's memory footprint increased for a considerable amount, mainly because of added caches for `Shoots`, `SecretBindings` and `Secrets`.

The required memory depends on the K8s cluster or the number/size of the stored resources, but we've observed increases from from ~20Mb to ~500Mb. Under consideration that such an admission component has multiple replicas and is only responsible for one provider, the runtime costs are too high.

Hence, we should try to:
1. Disable caching for all objects and analyse the consequences, i.e. is cost per admission request still acceptable
2. If 1.) is not feasible, use API Reader to read `Secrets` because we expect that `Secrets` are the most occurring resource kinds in the Garden cluster.

ialidzhikov:
Thank you @timuthy for opening this issue.

The avg memory usage now seem to be stable ~80Mb with spikes to 100Mb. These numbers on a landscape with ~10000 Secrets and ~7000 SecretBindings.

![Screenshot 2020-08-28 at 11 50 41](https://user-images.githubusercontent.com/9372594/91541755-7ee1a400-e925-11ea-8c1e-22f1b9f5e00f.png)

I can give a try for

> If 1.) is not feasible, use API Reader to read Secrets because we expect that Secrets are the most occurring resource kinds in the Garden cluster.

and see whether it improves the memory usage.

ialidzhikov:
I had a look today into this issue and tried to outline future improvements.

- I checked whether the admission-gcp component is using `application/vnd.kubernetes.protobuf` content type. According to the [protobuf proposal](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/api-machinery/protobuf.md), using protobuf can decrease the CPU, memory and network usage:
  > Experiments have demonstrated a 10x reduction in CPU use during serialization and deserialization, a 2x reduction in size in bytes on the wire, and a 6-9x reduction in the amount of objects created on the heap during serialization.

  But it seems that admission-gcp is already using the protobuf content type today (I think, by chance?) thanks to the `init()` in the `github.com/gardener/gardener/pkg/client/kubernetes` pkg that calls `apiutil.AddToProtobufScheme(protobufSchemeBuilder.AddToScheme)`:

[...]

PR #253, opened 2021-04-07
Use client.Reader to prevent starting Informer for Secrets

ialidzhikov:
/platform gcp

After https://github.com/gardener/gardener-extension-provider-gcp/pull/112 I think that we have only one place where we explicitly interact with Secrets. This PR replaces the usage of a cached client with a client.Reader for the corresponding call to prevent having an Informer for Secrets under the hood.

~~I hope that this will improve the memory usage of the admission-gcp component.~~
Update: see https://github.com/gardener/gardener-extension-provider-gcp/pull/253#issuecomment-814953825

Part of #143

**Release note**:

```other operator
admission-gcp memory usage is reduced by not using cache for reading Secrets.
```


ialidzhikov:
I deployed this change on a large landscape and it improved the memory usage.

Screenshot 2021-04-07 at 17 10 36

The memory usage dropped from `~ avg 185Mi` to `~ avg 118Mi`.

timebertt (approved):
/lgtm

PR #256, opened 2021-04-09
Switch to kaniko oci builder

ialidzhikov:
Similar to https://github.com/gardener/gardener-extension-provider-aws/pull/281

**Release note**:

```other operator
NONE
```

PR #257, opened 2021-04-09
depend on publish trait rather than injected step(s)

ccwienk:
prepare switching to kaniko-builder


ialidzhikov:
@ccwienk , can we close this one in favor of https://github.com/gardener/gardener-extension-provider-gcp/pull/256?

ialidzhikov (approved):
/lgtm

PR #296, opened 2021-07-16
Add validation for min # of worker machines in shoot

### LINKED ISSUE #295: Shoot worker validation

/area auto-scaling
/kind enhancement
/priority 3
/platform gcp

A validation in the shoot worker spec where the minimum number of workers in a worker pool is at least equal to the number of zones configured for that worker pool. 

**Why is this needed**:
When users set minimum worker to 1 in a multi zone worker pool of clusters it leads to autoscaler scale from zero issues which is not yet implemented for GCP. An example can be seen [here](https://sap-ti.slack.com/archives/CBV3JS9S4/p1626310288309100). This validation will help the user understand and configure the shoot correctly by themselves.

This validation can be removed once scale-from-zero is supported for GCP. The ongoing issue is [here](https://github.com/gardener/autoscaler/issues/27)

AxiomSamarth:

/area auto-scaling
/kind enhancement
/platform gcp

This PR adds validation in the shoot worker spec to check if the minimum number of workers in a worker pool is at least equal to the number of zones configured for that worker pool. This validation will be helpful for the user to figure out the problem until scale-from-zero is supported for GCP.

Fixes #295 

**Release note**:

```other user
Add validation for min # of workers required to support scale-up from zero. 
```

### DIFF
--- a/docs/usage-as-end-user.md
+++ b/docs/usage-as-end-user.md
@@ -125,7 +125,7 @@ If you don't want to configure anything for the `cloudControllerManager` simply 
 
 ## WorkerConfig
 
-Multiple zones can be configured for a worker group of a GCP Shoot. The minimum number of machines in every worker group should be equal to the number of zones configured for that worker-group.
+Multiple zones can be configured for a worker group of a GCP Shoot. The minimum number of machines in every worker group should be equal to or greater than the number of zones configured for that worker-group.
 
 > The predicate is - A worker group with N zones configured should have minimum N machines.
 
--- a/docs/usage-as-end-user.md
+++ b/docs/usage-as-end-user.md
@@ -125,6 +125,36 @@ If you don't want to configure anything for the `cloudControllerManager` simply 
 
 ## WorkerConfig
 
+Multiple zones can be configured for a worker group of a GCP Shoot. The minimum number of machines in every worker group should be equal to the number of zones configured for that worker-group.
+
+> The predicate is - A worker group with N zones configured should have minimum N machines.
+
+⚠️ This is important because, as of today, Cluster Autoscaler does not support scale-from-zero on GCP.
+
+The following YAML is a snippet of a `Shoot` resource:
+
+```
+workers:
+  - name: worker-vezh0
+    machine:
+      type: n1-standard-2
+      image:
+        name: gardenlinux
+        version: 318.8.0
+    maximum: 6
+    minimum: 2 # the value should be equal to or greater than the number of zones
+    maxSurge: 1
+    maxUnavailable: 0
+    volume:
+      type: pd-standard
+      size: 50Gi
+    zones:
+      - europe-west1-c
+      - europe-west1-d
+    systemComponents:
+      allow: true
+```
+
 The worker configuration contains:
 
 * Local SSD interface for the additional volumes attached to GCP worker machines.
@@ -147,7 +177,6 @@ serviceAccount:
   scopes:
   - https://www.googleapis.com/auth/cloud-platform
 ```
-
 ## Example `Shoot` manifest
 
 Please find below an example `Shoot` manifest:
--- a/pkg/admission/validator/shoot.go
+++ b/pkg/admission/validator/shoot.go
@@ -159,6 +159,14 @@ func (s *shoot) validateCreate(ctx context.Context, shoot *core.Shoot) error {
 		return err
 	}
 
+	// TODO: This check won't be needed after generic support to scale from zero is introduced in CA
+	// Ongoing issue - https://github.com/gardener/autoscaler/issues/27
+	for i, worker := range shoot.Spec.Provider.Workers {
+		if worker.Minimum < int32(len(worker.Zones)) {
+			return fmt.Errorf("%s minimum value must be >= %d if maximum value > 0 (auto scaling to 0 & from 0 is not supported", workersPath.Index(i).Child("minimum").String(), len(worker.Zones))
+		}
+	}
+
 	if err := s.validateContext(validationContext).ToAggregate(); err != nil {
 		return err
 	}
--- a/pkg/apis/gcp/validation/shoot.go
+++ b/pkg/apis/gcp/validation/shoot.go
@@ -15,6 +15,8 @@
 package validation
 
 import (
+	"fmt"
+
 	"github.com/gardener/gardener/pkg/apis/core"
 	"github.com/gardener/gardener/pkg/apis/core/validation"
 
@@ -73,13 +75,19 @@ func validateVolume(vol *core.Volume, fldPath *field.Path) field.ErrorList {
 func ValidateWorkersUpdate(oldWorkers, newWorkers []core.Worker, fldPath *field.Path) field.ErrorList {
 	allErrs := field.ErrorList{}
 	for i, newWorker := range newWorkers {
+
+		workerFldPath := fldPath.Index(i)
 		for _, oldWorker := range oldWorkers {
 			if newWorker.Name == oldWorker.Name {
 				if validation.ShouldEnforceImmutability(newWorker.Zones, oldWorker.Zones) {
-					allErrs = append(allErrs, apivalidation.ValidateImmutableField(newWorker.Zones, oldWorker.Zones, fldPath.Index(i).Child("zones"))...)
+					allErrs = append(allErrs, apivalidation.ValidateImmutableField(newWorker.Zones, oldWorker.Zones, workerFldPath.Child("zones"))...)
 				}
 				break
 			}
+		}
+
+		if newWorker.Minimum < int32(len(newWorker.Zones)) {
+			allErrs = append(allErrs, field.Forbidden(workerFldPath.Child("minimum"), fmt.Sprintf("minimum value must be >= %d if maximum value > 0 (auto scaling to 0 & from 0 is not supported", len(newWorker.Zones))))
 		}
 	}
 	return allErrs
--- a/pkg/apis/gcp/validation/shoot_test.go
+++ b/pkg/apis/gcp/validation/shoot_test.go
@@ -15,6 +15,8 @@
 package validation_test
 
 import (
+	"fmt"
+
 	"github.com/gardener/gardener-extension-provider-gcp/pkg/apis/gcp"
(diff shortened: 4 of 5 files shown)


prashanth26 (approved):
Thanks for these changes @AxiomSamarth . 

/lgtm
/needs second-opinion

rfranzke:
Is this a backwards-incompatible change? I assume my `Shoot` will not be able to reconcile anymore after this is rolled out, right? Shouldn't we make this validation more "graceful" (e.g., only for new shoots or newly added worker pools)?
/cc @ialidzhikov @vpnachev

vpnachev:
I agree with @rfranzke, this change should be more graceful without breaking existing clusters.

prashanth26:
> Is this a backwards-incompatible change? I assume my `Shoot` will not be able to reconcile anymore after this is rolled out, right? Shouldn't we make this validation more "graceful" (e.g., only for new shoots or newly added worker pools)?
> /cc @ialidzhikov @vpnachev

I guess the shoot's that will fail to reconcile are the only ones where the workers have been wrongly configured (min is less the number of zones) right? So I think if such shoot reconciliations are failing, they would be fixed once the owner fixes the shoot and the shoot will continue to exist however reconciliation would fail until then? If this is the case, I would think it's the right thing.

Also, if you guys aren't sure about this validation check (considering enabling for only new shoots/workers), we could probably just document this for now? and anyways when the genric scale to zero support is in, we will anyways get rid of these checks/doc?

AxiomSamarth:
As it was also concurred in today's DoD Sync, this validation is good to have here.

@rfranzke However, I have a doubt in implementation

>validate only for new shoots or newly added worker pools

I know that I can refer to `generation` field in the Shoot `metadata` to understand if this is a new Shoot. But how do I figure out if a new worker pool is added ?

dkistner:
You could do the respective check also in the `ValidateWorkersUpdate()`. This gets a list of the new and old worker pools. With that you should be able to detect this.

pkg/apis/gcp/validation/shoot_test.go
```diff
@@ -79,3 +144,7 @@ func validateWorkerConfig(workers []core.Worker, workerConfig *gcp.WorkerConfig)
 
 	return allErrs
 }
+
+func getStringPointer(s string) *string {
+	return &s
+}
```
kon-angelo, line 150: You can use something like `pointer.StringPtr()` like we use [here](https://github.com/gardener/gardener-extension-provider-gcp/blob/7074d6c4528d46e1c72e5200ecf65b2ff869c67b/pkg/apis/gcp/validation/infrastructure_test.go#L58).

AxiomSamarth:
@dkistner I have added the checks in `ValidateWorkerUpdates()` function to validate new worker pools and also a similar check in the `ValidateCreate()` to validate new shoots. Thus, validating this gracefully as @rfranzke mentioned. I have also added a couple of Unit Tests for the same.

Let me know if this change is as expected.

pkg/admission/validator/shoot.go
```diff
@@ -159,6 +159,17 @@ func (s *shoot) validateCreate(ctx context.Context, shoot *core.Shoot) error {
 		return err
 	}
 
+	// TODO: This check won't be needed after generic support to scale from zero is introduced in CA
+	// Ongoing issue - https://github.com/gardener/autoscaler/issues/27
+	workers := shoot.Spec.Provider.Workers
+	for i, worker := range workers {
```
dkistner, line 165: Nit 
```suggestion
	for i, worker := range shoot.Spec.Provider.Workers {
```

pkg/admission/validator/shoot.go
```diff
@@ -159,6 +159,17 @@ func (s *shoot) validateCreate(ctx context.Context, shoot *core.Shoot) error {
...
+		workerFldPath := workersPath.Index(i)
+		if worker.Maximum != 0 && worker.Minimum < int32(len(worker.Zones)) {
```
dkistner, line 167: Why is the `worker.Maximum != 0` combined with the `worker.Minimum < int32(len(worker.Zones))` check?

I mean `worker.Maximum` needs always be greater than 0 (until we have scale to zero for gcp) and greater equal `worker.Minimum`, right? So should `worker.Maximum == 0` not be an independent check?

pkg/apis/gcp/validation/shoot.go
```diff
@@ -81,6 +85,10 @@ func ValidateWorkersUpdate(oldWorkers, newWorkers []core.Worker, fldPath *field.
 				break
 			}
 		}
+
+		if newWorker.Maximum != 0 && newWorker.Minimum < int32(len(newWorker.Zones)) {
```
dkistner, line 89: Same as above.

pkg/admission/validator/shoot.go
AxiomSamarth: You are right, @dkistner. The `worker.maximum` check need not be here. There is already a validation present [here](https://github.com/gardener/gardener-extension-provider-gcp/blob/21db727e7c36db6a5cf549e789c59c46fbb61683/pkg/apis/gcp/validation/shoot.go#L53-L55) for the `worker.Maximum` value. I shall update the code.

docs/usage-as-end-user.md
```diff
@@ -125,6 +125,36 @@ If you don't want to configure anything for the `cloudControllerManager` simply
...
+```
+workers:
+  - name: worker-vezh0
+    machine:
+      type: n1-standard-2
+      image:
+        name: gardenlinux
+        version: 318.8.0
+    maximum: 6
+    minimum: 2 # the value should be equal to the number of zones
```
vpnachev, line 145:
```suggestion
    minimum: 2 # the value should be equal or greater than the number of zones
```

docs/usage-as-end-user.md
```diff
@@ -125,6 +125,36 @@ If you don't want to configure anything for the `cloudControllerManager` simply
 
 ## WorkerConfig
 
+Multiple zones can be configured for a worker group of a GCP Shoot. The minimum number of machines in every worker group should be equal to the number of zones configured for that worker-group.
```
vpnachev, line 128: Please trim the trailing white spaces.
```suggestion
Multiple zones can be configured for a worker group of a GCP Shoot. The minimum number of machines in every worker group should be equal to the number of zones configured for that worker-group.
```

pkg/admission/validator/shoot.go
```diff
@@ -159,6 +159,16 @@ func (s *shoot) validateCreate(ctx context.Context, shoot *core.Shoot) error {
 		return err
 	}
 
+	// TODO: This check won't be needed after generic support to scale from zero is introduced in CA
+	// Ongoing issue - https://github.com/gardener/autoscaler/issues/27
+	for i, worker := range shoot.Spec.Provider.Workers {
+		workerFldPath := workersPath.Index(i)
+		if worker.Minimum < int32(len(worker.Zones)) {
+			err := fmt.Errorf(workerFldPath.Child("minimum").String() + " minimum value must be >= " + fmt.Sprint(len(worker.Zones)) + " if maximum value > 0 (auto scaling to 0 & from 0 is not supported")
+			return err
```
vpnachev, line 168:
```suggestion
			return fmt.Errorf("%s minimum value must be >= %d if maximum value > 0 (auto scaling to 0 & from 0 is not supported", workerFldPath.Child("minimum").String(), len(worker.Zones))
```

pkg/apis/gcp/validation/shoot.go
```diff
@@ -81,6 +85,10 @@ func ValidateWorkersUpdate(oldWorkers, newWorkers []core.Worker, fldPath *field.
 				break
 			}
 		}
+
+		if newWorker.Minimum < int32(len(newWorker.Zones)) {
+			allErrs = append(allErrs, field.Forbidden(workerFldPath.Child("minimum"), "minimum value must be >= "+fmt.Sprint(len(newWorker.Zones))+" if maximum value > 0 (auto scaling to 0 & from 0 is not supported"))
```
vpnachev, line 90:
```suggestion
			allErrs = append(allErrs, field.Forbidden(workerFldPath.Child("minimum"), fmt.Sprintf("minimum value must be >= %d if maximum value > 0 (auto scaling to 0 & from 0 is not supported", len(newWorker.Zones))))
```

pkg/apis/gcp/validation/shoot.go
```diff
@@ -73,6 +75,8 @@ func validateVolume(vol *core.Volume, fldPath *field.Path) field.ErrorList {
 func ValidateWorkersUpdate(oldWorkers, newWorkers []core.Worker, fldPath *field.Path) field.ErrorList {
 	allErrs := field.ErrorList{}
 	for i, newWorker := range newWorkers {
+
+		workerFldPath := fldPath.Index(i)
```
vpnachev, line 79: You can use this field path also on line 83.

vpnachev (commented):
Some code style suggestions.

AxiomSamarth:
> Some code style suggestions.

Thank you @vpnachev. Those were very good suggestions.

docs/usage-as-end-user.md
vpnachev: Why was this labeled as resolved when it is still not?

vpnachev: Ok, I messed the things, but this can be applied also to line 128
> Multiple zones can be configured for a worker group of a GCP Shoot. The minimum number of machines in every worker group should be equal to the number of zones configured for that worker-group.

docs/usage-as-end-user.md
```diff
@@ -125,6 +125,36 @@ If you don't want to configure anything for the `cloudControllerManager` simply
...
```
vpnachev, line 128:
```suggestion
Multiple zones can be configured for a worker group of a GCP Shoot. The minimum number of machines in every worker group should be equal or grater than the number of zones configured for that worker-group.
```

vpnachev (commented):
Ovelall looks good to me, just one nit.

vpnachev:
The release is done and available https://github.com/gardener/gardener-extension-provider-gcp/releases/tag/v1.18.0
/unhold

vpnachev (approved):
/lgtm

prashanth26 (approved):
/lgtm

PR #306, opened 2021-08-11
Allow to enabled ccm alpha features

dkistner:

/area open-source
/kind enhancement
/platform gcp

Allow to enable cloud controller manager alpha features. 
Alpha features can only be enabled when the `AllAlpha` feature flag is activated.

End to end test is missing therefore I keep it in draft state.

**Release note**:

```feature user
Allow to enable cloud-controller-manager alpha features via `cloudControllerManager.alphaFeatures[]` in the `ControlPlaneConfig. Alpha features can only be enabled when the `AllAlpha` feature flag is active.
```

/invite @prashanth26 @DockToFuture

### DIFF
--- a/charts/internal/cloud-provider-config/templates/cloud-provider-config.yaml
+++ b/charts/internal/cloud-provider-config/templates/cloud-provider-config.yaml
@@ -15,3 +15,6 @@ data:
     local-zone="{{ .Values.zone }}"
     token-url=nil
     node-tags="{{ .Values.nodeTags }}"
+    {{- if .Values.alphaFeatures }}
+    alpha-features="{{range $index, $feature := .Values.alphaFeatures}}{{if $index}},{{end}}{{$feature}}{{end}}"
+    {{- end }}
--- a/docs/usage-as-end-user.md
+++ b/docs/usage-as-end-user.md
@@ -123,6 +123,9 @@ The `cloudControllerManager.featureGates` contains a map of explicitly enabled o
 For production usage it's not recommend to use this field at all as you can enable alpha features or disable beta/stable features, potentially impacting the cluster stability.
 If you don't want to configure anything for the `cloudControllerManager` simply omit the key in the YAML specification.
 
+The `cloudControllerManager.alphaFeatureGates` contains a map of enabled or disabled feature gates.
+The alpha features can only be enabled in when the `AllAlpha` feature gate is enabled.
+
 ## WorkerConfig
 
 Multiple zones can be configured for a worker group of a GCP Shoot. The minimum number of machines in every worker group should be equal to or greater than the number of zones configured for that worker-group.
--- a/pkg/apis/gcp/v1alpha1/types_controlplane.go
+++ b/pkg/apis/gcp/v1alpha1/types_controlplane.go
@@ -38,4 +38,7 @@ type CloudControllerManagerConfig struct {
 	// FeatureGates contains information about enabled feature gates.
 	// +optional
 	FeatureGates map[string]bool `json:"featureGates,omitempty"`
+	// AlphaFeatureGates contains information about enabled alpha feature gates.
+	// Alpha Features require the "AllAlpha" feature gate enabled on the Cloud Controller Manager.
+	AlphaFeatureGates map[string]bool `json:"alphaFeatureGates,omitempty"`
 }
--- a/pkg/controller/controlplane/valuesprovider.go
+++ b/pkg/controller/controlplane/valuesprovider.go
@@ -417,7 +417,22 @@ func getConfigChartValues(
 		"subNetworkName": subNetworkName,
 		"zone":           cpConfig.Zone,
 		"nodeTags":       cp.Namespace,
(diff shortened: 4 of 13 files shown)


vpnachev (commented):
Could you implement some more enhanced validation like https://github.com/gardener/gardener/pull/4149? This way it will not be possible for the shoot owners to specify invalid feature gates that may prevent the CCM to start-up.

And in general, so far we never enabled alpha features because they can be removed at any time which makes the effort to enable it obsolete, but increases the operational burden.

dkistner:
Generally I agree with you to not enable alpha features. I'm currently also not sure if we still need this PR as we had one stakeholder which require the `ILBCustomSubnet` because they need internal loadbalancers with ips from a specific subnet.
As the ccm in `v1.21` supports this without this alpha feature enabled and they start to adopt k8s `v1.21` we can probably also close this. Just removed the draft for completeness.
@stiller-leser Can you confirm?

pkg/apis/gcp/v1alpha1/types_controlplane.go
```diff
@@ -38,4 +38,7 @@ type CloudControllerManagerConfig struct {
 	// FeatureGates contains information about enabled feature gates.
 	// +optional
 	FeatureGates map[string]bool `json:"featureGates,omitempty"`
+	// AlphaFeatureGates contains information about enabled alpha feature gates.
+	// Alpha Features require the "AllAlpha" feature gate enabled on the Cloud Controller Manager.
```
rfranzke, line 42:
```suggestion
	// Alpha Features require the "AllAlpha" feature gate enabled on the Cloud Controller Manager.
	// +optional
```

pkg/controller/controlplane/valuesprovider.go
```diff
@@ -417,9 +417,24 @@ func getConfigChartValues(
 		"subNetworkName": subNetworkName,
 		"zone":           cpConfig.Zone,
 		"nodeTags":       cp.Namespace,
+		"alphaFeatures":  getAlphaFeatuesValues(cpConfig.CloudControllerManager),
 	}, nil
 }
 
+func getAlphaFeatuesValues(ccmConfig *apisgcp.CloudControllerManagerConfig) []string {
+	var alphaFeatues = []string{}
```
rfranzke, line 425:
```suggestion
	var alphaFeatues []string
```

pkg/controller/controlplane/valuesprovider.go
```diff
@@ -417,9 +417,24 @@ func getConfigChartValues(
...
+
+	if ccmConfig == nil {
+		return alphaFeatues
+	}
```
rfranzke, line 429:
```suggestion
	if ccmConfig == nil {
		return nil
	}
```

and move to the top?

rfranzke (commented):
Minor feedback in case the PR is still relevant

stiller-leser:
Hi all,

We do not need it anymore, but since it is almost done, wouldn't it be a useful addition?

Best regards

________________________________

charts/internal/cloud-provider-config/templates/cloud-provider-config.yaml
```diff
@@ -15,3 +15,6 @@ data:
     local-zone="{{ .Values.zone }}"
     token-url=nil
     node-tags="{{ .Values.nodeTags }}"
+    {{- if .Values.alphaFeatures }}
+    alpha-features="{{range $index, $feature := .Values.alphaFeatures}}{{if $index}},{{end}}{{$feature}}{{end}}"
```
ialidzhikov, line 19: Ha, I didn't know that gcp CCM supports such config option. It is confusing and not what I would expect honestly. CCMs have a `--feature-gates` flag and I would expect even out-of-tree provider-specific features to be configured via this flag.

docs/usage-as-end-user.md
```diff
@@ -123,6 +123,9 @@ The `cloudControllerManager.featureGates` contains a map of explicitly enabled o
 For production usage it's not recommend to use this field at all as you can enable alpha features or disable beta/stable features, potentially impacting the cluster stability.
 If you don't want to configure anything for the `cloudControllerManager` simply omit the key in the YAML specification.
 
+The `cloudControllerManager.alphaFeatureGates` contains a map of enabled or disabled feature gates.
+The alpha features can only be enabled in when the `AllAlpha` feature gate is enabled.
```
ialidzhikov, line 127: Isn't `AllAlpha` too dangerous? I guess currently there are alpha features that can cause issues when they are enabled. One good example is the `MemoryQoS` feature gate that requires (I expect) cgroup v2 that is currently not supported (disabled) by gardenlinux - https://github.com/gardenlinux/gardenlinux/pull/205.
I guess no one of us would like to dig for issue in a cluster where AllAlpha feature gate is enabled.

dkistner:
As said, this was intended to unblock @stiller-leser use case, but as they now have a solution by with moving their clusters to k8s `v1.21` that does not require the alpha feature anymore I would also tend to not allowing enabling alpha features.
So I would close this PR. In case we need it we can reactive again.
/close

PR #312, opened 2021-08-25
Add infrastructure ConfigValidator for checking NAT IPs

### LINKED ISSUE #309: Classify missing CloudNAT IPs as configuration error

/area ops-productivity
/area quality
/kind bug
/priority 3
/platform gcp

**What happened**:
Infrastructure reconciliation does not check whether the given CloudNAT IP exists. In such case it yields a terraform error that is not intuitive.

provider-gcp to perform a prerequisite check whether the given CloudNAT IP exist. In such case provider-gcp should return an appropriate error and should add the `ERR_CONFIGURATION_PROBLEM` error code.

1. Create a Shoot with a CloudNAT IP that does not exist

```yaml
kind: InfrastructureConfig
networks:
  workers: 10.250.0.0/16
cloudNAT:
   minPortsPerVM: 2048
  natIPNames:
  - name: missingone
```

2. Make sure that the Infrastructure reconciliation fails with:

```
 * Null value found in list
on tf/main.tf line 42, in resource "google_compute_router_nat" "nat":
42:   nat_ips                = [data.google_compute_address.<address_name>.self_link]
Null values are not allowed for this attribute value. 
```

**Environment**:

- Extension version: v1.18.0
- Kubernetes version (use `kubectl version`):
- Others:

stoyanr:

/area ops-productivity
/area quality
/kind bug
/platform gcp

* Adds an implementation of the `infrastructure.ConfigValidator` for checking whether NAT IP names refer to existing and available external IP addresses. See https://github.com/gardener/gardener/pull/4547.
* Adds unit tests for the `infrastructure.ConfigValidator` implementation.
* Vendors g/g v1.31.0-dev (in a separate commit).

Fixes #309

**Release note**:

```feature operator
NAT IP names in infrastructure config are now checked if they exist and are available, and if not the issue is reported as `ERR_CONFIGURATION_PROBLEM` with a clear error message.
```


pkg/controller/infrastructure/configvalidator.go
```diff
@@ -0,0 +1,102 @@
...
+	// Create GCP compute client
+	computeClient, err := c.gcpClientFactory.NewComputeClient(ctx, c.Client(), infra.Spec.SecretRef)
+	if err != nil {
+		allErrs = append(allErrs, field.InternalError(nil, err))
+		return allErrs
+	}
+
+	// Validate infrastructure config
+	switch {
+	case config.Networks.CloudNAT != nil:
```
kon-angelo, line 69: Small nit: not a big fan of switch here because if we were to add extra checks in the switch we would have to follow with a `fallthrough` which arguably could be easy to miss.

pkg/controller/infrastructure/configvalidator.go
```diff
@@ -0,0 +1,102 @@
...
+	externalAddresses, err := computeClient.GetExternalAddresses(ctx, region)
+	if err != nil {
+		allErrs = append(allErrs, field.InternalError(fldPath, err))
+		return allErrs
+	}
+
+	// Check whether each specified NAT IP name exists and is available
+	for i, natIP := range cloudNAT.NatIPNames {
+		natIPNamePath := fldPath.Child("napIPNames").Index(i).Child("name")
+		if available, ok := externalAddresses[natIP.Name]; !ok {
```
kon-angelo, line 94: The check for `GetExternalAddresses` returned map is:

```
addresses[address.Name] = address.Status == "RESERVED"
```

Am I wrong then in thinking that if the value returned is _true_ then the nat ip is _unavailable_ ? If yes, that would mean we have to fix the name or the respective condition.

stoyanr: `true` here means the nat ip is *available* as originally intended, see https://pkg.go.dev/google.golang.org/api@v0.54.0/compute/v1#Address:

```
	// Status: [Output Only] The status of the address, which can be one of
	// RESERVING, RESERVED, or IN_USE. An address that is RESERVING is
	// currently in the process of being reserved. A RESERVED address is
	// currently reserved and available to use. An IN_USE address is
	// currently being used by another resource and is not available.
```

stoyanr: Also, I have of course tested the above implementation and verified that it works as intended, including for IP addresses that are not available.

kon-angelo: @stoyanr thank you for the explanation. That makes sense .

pkg/controller/infrastructure/configvalidator.go
stoyanr: Yes, this `switch` is rather misplaced, changed to `if`.

stoyanr:
@kon-angelo Thanks for your comments, addressed them and also added some unit tests.

pkg/gcp/client/compute.go
```diff
@@ -0,0 +1,81 @@
...
+		return nil, err
+	}
+
+	return &computeClient{
+		service:   service,
+		projectID: credentials.ProjectID,
+	}, nil
+}
+
+func NewComputeClientFromSecretRef(ctx context.Context, c client.Client, secretRef corev1.SecretReference) (ComputeClient, error) {
```
rfranzke, line 58: Nit: doc string missing

rfranzke (approved):
/lgtm

dkistner (dismissed):
/lgtm

pkg/gcp/client/compute.go
stoyanr: Done, also for other similar methods with missing doc strings.
I wonder, should we rather make the linter fail on missing doc strings? I am a bit tired of fixing such "nits", not to mention that they are often missed (leading to more "nits" after copy and paste).

rfranzke (approved):

PR #319, opened 2021-08-31
Make the linter fail on missing or wrong doc strings

stoyanr:

/area quality, dev-productivity
/kind enhancement
/platform gcp

Makes the linter fail on missing or wrong doc comments and fixes all such issues that already exist.

Fixes #

* Making the linter fail on missing or wrong doc comments consists of enabling `golint` and excluding certain `golint` errors that we are perhaps not interested in. We could perhaps keep some of these (everything besides "don't use underscores in Go names" is a candidate IMO), there are not too many such errors yet.
* This PR is intended as a "demo", if we agree it makes sense similar PRs could be opened in g/g and other extensions.

**Release note**:

```other developer
Missing or wrong doc comments will now be reported by the linter.
```

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -5,3 +5,11 @@ run:
 linters:
   disable:
   - unused
+  enable:
+  - golint
+
+issues:
+  exclude-use-default: false
+  exclude:
+  # golint:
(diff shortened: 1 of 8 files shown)


rfranzke (approved):
/lgtm

PR #320, opened 2021-09-01
Update kubernetes-csi/external-snapshotter to v4.2.1

yuvalman:
/area storage
/kind enhancement
/platform gcp

Upgrade external-snapshotter to v4 is a must for being aligned with external-provisioner v3.0.0:
https://github.com/kubernetes-csi/external-provisioner/blob/release-3.0/CHANGELOG/CHANGELOG-3.0.md#no-really-you-must-read-this-before-you-upgrade 

We have a csi-provision driver that fails because of this issue:
https://github.com/kubernetes-csi/external-provisioner/issues/582

**Release note**:
```other operator
The following images are updated:
- k8s.gcr.io/sig-storage/csi-snapshotter: v3.0.3 -> v4.2.1
- k8s.gcr.io/sig-storage/snapshot-controller: v3.0.3 -> v4.2.1

Volume snapshots CRD version was updated to v1

```

### DIFF
--- a/charts/images.yaml
+++ b/charts/images.yaml
@@ -50,6 +50,12 @@ images:
   sourceRepository: github.com/kubernetes-csi/external-provisioner
   repository: k8s.gcr.io/sig-storage/csi-provisioner
   tag: "v2.1.2"
+  targetVersion: "< 1.20"
+- name: csi-provisioner
+  sourceRepository: github.com/kubernetes-csi/external-provisioner
+  repository: k8s.gcr.io/sig-storage/csi-provisioner
+  tag: "v3.0.0"
+  targetVersion: ">= 1.20"
 - name: csi-attacher
   sourceRepository: github.com/kubernetes-csi/external-attacher
   repository: k8s.gcr.io/sig-storage/csi-attacher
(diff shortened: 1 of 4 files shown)


prashanth26 (commented):
This PR also has to be adopted W.R.T [this comment](https://github.com/gardener/gardener-extension-provider-gcp/pull/321/files#r704090886). Also was wondering if there is a reason why the two PRs are seperate? Can it be combined into a single change?

yuvalman:
Hi @prashanth26, The reason I separated the prs is that it upgrades different components, if we would like to change/revert only one of them, it will be easier to do that because of the git history. 
We can combine them if you think it's better in this case (I added a new commit for that).

Please let me know what do you think.

prashanth26:
> Hi @prashanth26, The reason I separated the prs is that it upgrades different components, if we would like to change/revert only one of them, it will be easier to do that because of the git history.
> We can combine them if you think it's better in this case (I added a new commit for that).
> 
> Please let me know what do you think.

Sure thanks. I think a single PR should suffice. Also, we would still need checks to apply this only on the newer versions of k8s for the other components as well right? Unsure how the other file changes would affect the older k8s versions though.

PR #331, opened 2021-09-30
Fix config validation of NAT IP addresses

stoyanr:

/area networking
/kind bug
/platform gcp

Fixes config validation of NAT IP addresses to correctly recognise if the external IP address is in use by the shoot's router.

Fixes #

**Release note**:

```bugfix user
The cloud NAT IP validation has been fixed to correctly recognise if the external IP address is in use by the shoot's router.
```

### DIFF
--- a/pkg/controller/infrastructure/configvalidator.go
+++ b/pkg/controller/infrastructure/configvalidator.go
@@ -17,6 +17,7 @@ package infrastructure
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	api "github.com/gardener/gardener-extension-provider-gcp/pkg/apis/gcp"
 	"github.com/gardener/gardener-extension-provider-gcp/pkg/apis/gcp/helper"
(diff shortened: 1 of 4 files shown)


pkg/controller/infrastructure/configvalidator.go
```diff
@@ -66,35 +67,39 @@ func (c *configValidator) Validate(ctx context.Context, infra *extensionsv1alpha
...
-	for i, natIP := range cloudNAT.NatIPNames {
-		natIPNamePath := fldPath.Child("natIPNames").Index(i).Child("name")
-		if available, ok := externalAddresses[natIP.Name]; !ok {
+	for i, natIP := range networks.CloudNAT.NatIPNames {
+		natIPNamePath := fldPath.Child("cloudNAT", "natIPNames").Index(i).Child("name")
+		if userNames, ok := externalAddresses[natIP.Name]; !ok {
 			allErrs = append(allErrs, field.NotFound(natIPNamePath, natIP.Name))
-		} else if !available {
-			allErrs = append(allErrs, field.Invalid(natIPNamePath, natIP.Name, "external IP address is already in use"))
+		} else if len(userNames) > 1 || len(userNames) == 1 && userNames[0] != cloudRouterName {
```
kon-angelo, line 100: Optional:
```
else if len(userNames) != 1 || userNames[0] != cloudRouterName
```

kon-angelo (approved):
/lgtm

ialidzhikov (approved):
Thanks! I locally managed to reproduce the issue with `provider-gcp@v1.19.0` - about wrong validation that does not consider the case when the IP address is IN_USE by the current cloud route. I also verified that this PR is fixing this case.

/lgtm
/needs cherry-pick

PR #336, opened 2021-10-12
Slightly clarify shoot.worker validation error

plkokanov:

/area ops-productivity
/kind enhancement
/platform gcp

Clarifies the validation error message when specifying a value for `shoot.spec.providers.workers[].minimum`  that is less than the number of `shoot.spec.providers.workers[].zones` 

```
Forbidden: spec.provider.workers[0].minimum value must be >= 2 (number of zones) if maximum value > 0 (auto scaling to 0 & from 0 is not supported)
```
 
None

I've added an additional check to only return the error when `workers.Maximum > 0` as per the text that was already present in the error.

**Release note**:

```other operator
Clarify shoot validation error when the `shoot.spec.provider.workers[].Minimum` value is less than the number of `shoot.spec.provider.workers[].zones`
```

### DIFF
--- a/pkg/apis/gcp/validation/shoot.go
+++ b/pkg/apis/gcp/validation/shoot.go
@@ -60,6 +60,15 @@ func ValidateWorkers(workers []core.Worker, fldPath *field.Path) field.ErrorList
 	return allErrs
 }
 
+// ValidateWorkerAutoScaling checks if the worker.minimum value is greater or equal to the number of worker.zones[]
+// when the worker.maximum value is greater than zero. This check is necessary because autoscaling from 0 is not supported on gcp.
+func ValidateWorkerAutoScaling(worker core.Worker, path string) error {
+	if worker.Maximum > 0 && worker.Minimum < int32(len(worker.Zones)) {
+		return fmt.Errorf("%s value must be >= %d (number of zones) if maximum value > 0 (auto scaling to 0 & from 0 is not supported)", path, len(worker.Zones))
+	}
+	return nil
+}
+
 func validateVolume(vol *core.Volume, fldPath *field.Path) field.ErrorList {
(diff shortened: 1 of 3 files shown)


pkg/apis/gcp/validation/shoot.go
```diff
@@ -60,6 +60,15 @@ func ValidateWorkers(workers []core.Worker, fldPath *field.Path) field.ErrorList
 	return allErrs
 }
 
+// ValidateWorkerAutoScaling checks if the worker.minimum value is greater or equal to the number of worker.zones[]
+// when the worker.maximum value is greater than zero. This check is necessary because autoscaling from 0 is not supported on gcp.
+func ValidateWorkerAutoScaling(worker core.Worker, path string) error {
```
rfranzke, line 65: Why is this function exported/public?

plkokanov: Made it public so that i can use it in the admission package as well:
```
if worker.Minimum < int32(len(worker.Zones)) {
			return fmt.Errorf("%s minimum value must be >= %d if maximum value > 0 (auto scaling to 0 & from 0 is not supported", workersPath.Index(i).Child("minimum").String(), len(worker.Zones))
}
```

was called directly in `validateCreate()` in the admission package and I don't know why it wasn't part of the `ValidateWorkers()` method from the validator package.
Also I don't know why the worker config is generally validated in multiple places, check [this](https://github.com/gardener/gardener-extension-provider-gcp/blob/8b6696e0db54d34993662cfb34bbca4401d60238/pkg/admission/validator/shoot.go#L137) and [this](https://github.com/gardener/gardener-extension-provider-gcp/blob/8b6696e0db54d34993662cfb34bbca4401d60238/pkg/admission/validator/shoot.go#L141-L151) code.

PR #350, opened 2021-11-12
Enable specifying prefixed zone names in `DNSRecord` `spec.zone`

stoyanr:

/area control-plane
/kind enhancement
/platform gcp

Enables specifying a zone name prefixed with the project id, in addition to an unprefixed zone name, in the `spec.zone` field of a `DNSRecord` resource. This is needed to ensure that explicitly configured zones in landscapes and shoot `spec.dns.providers` configurations work for both the `DNSRecord` controller and [external-dns-management](https://github.com/gardener/external-dns-management), which always uses prefixed zone names on GCP.

Fixes #

* In order to avoid "overlapping zones" errors from `external-dns-management`, it is important to be able to configure zones explicitly in a way that works for both `external-dns-management` and the `DNSRecord` controller. Since the handling of zones in `external-dns-management` is harder to change, this PR proposes adapting the `DNSRecord` controller to the zone identification used by `external-dns-management`.

**Release note**:

```feature operator
It is now possible to specify a zone name prefixed with the project id, in addition to an unprefixed zone name, in the `spec.zone` field of a `DNSRecord` resource.
```

### DIFF
--- a/pkg/gcp/client/dns.go
+++ b/pkg/gcp/client/dns.go
@@ -20,6 +20,7 @@ import (
 	"strings"
 
 	"github.com/gardener/gardener-extension-provider-gcp/pkg/gcp"
+
 	"golang.org/x/oauth2"
 	"golang.org/x/oauth2/google"
 	googledns "google.golang.org/api/dns/v1"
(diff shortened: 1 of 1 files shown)


MartinWeindel (dismissed):
/lgtm

MartinWeindel (approved):

dkistner (approved):
/lgtm

PR #365, opened 2021-12-20
Support for service accounts managed by Gardener operators

### LINKED ISSUE #362: Implement a default centrally managed GCP service account for authentication to GCP API across GCP projects

/area security
/kind enhancement
/priority 2
/platform gcp

Allow Gardener service providers to offer a default GCP service account for authentication to the GCP APIs for managing resources inside different GCP projects that belong to one and the same GCP organization 
**Why is this needed**:
When Gardener service consumers want to start using Gardener to manage shoot clusters on GCP, they have to create in their GCP project a service account with a key and to grant permissions for compute administration and service accounts administration and usage to this GCP service account. Then they have to create a secret in their Gardener project where to configure the details about the GCP project, service account and the key. This service account key is then used by Gardener for authentication to the GCP API when managing GCP project resources as part of the Kubernetes cluster management activities. GCP service account keys are static secrets and their owner should make sure that they are rotated properly. 

When Gardener is hosted for internal corporate purposes, it is possible to **configure one central "Gardener" GCP service account in a separate GCP project (owned by Gardener ops team)** of the GCP organization. The key for this service account will be managed (rotated) by the Gardener ops team as the owner of the GCP service account and consumers will have to just grant the necessary permissions to this central service account in their projects, so that the Gardener service will have permissions to manage resources in their project using this service account key for authentication. This proposal is similar to the one made for an [Azure default Gardener SPN per Azure AD](https://github.com/gardener/gardener-extension-provider-azure/issues/383)

More details about how to use GCP service accounts across GCP projects are available here: [Centralize service accounts in separate projects.](https://cloud.google.com/iam/docs/service-accounts#locations)

dkistner:
/area security
/kind enhancement
/platform gcp

Add support to use a service account managed by Gardener operators for a GCP Shoot.

This eliminates the need for users to bring their own service account.
The service account can be managed by the Gardener operator in a different project in the same GCP organisation.
Users just need to assign the managed service account to their own project with proper permissions.
This gives the operator also the ability to rotate credentials in centralised way.

Fixes #362

**Release note**:

```feature operator
A feature has been added that allow to use a service account for a GCP Shoot that is managed by the Gardener operators. This eliminates the need for users for provide and manage the service account on his/her own. Find more information how to use this feature in the [end user documentation](https://github.com/gardener/gardener-extension-provider-gcp/blob/master/docs/usage-as-end-user.md#managed-service-accounts)
```

Draft until a few pending tests are completed.

/squash
/cc @donistz


pkg/webhook/cloudprovider/ensurer.go
```diff
@@ -0,0 +1,130 @@
+// Copyright (c) 2021 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
kon-angelo, line 1: Small nit but I think copyrights should be 2022

PR #364, opened 2021-12-16
Revendor `gardener/gardener@v1.38` and update Golang version to `1.17`

rfranzke:
**How to categorize this PR?**

/area open-source
/kind enhancement
/platform gcp

**What this PR does / why we need it**:
Revendor `gardener/gardener@v1.38` and update Golang version to `1.17`.

ℹ️ The `branch.cfg` definition is still using Golang `1.16`, hence, the `verify` step fails. Once this PR is approved, we can adapt the `branch.cfg` (also update to `1.17`) and then re-trigger the CI pipeline for this PR.

**Which issue(s) this PR fixes**:
Part of https://github.com/gardener/gardener/issues/4866

**Release note**:

```noteworthy developer
The Golang version has been updated to `1.17.5`.
```

PR #368, opened 2021-12-22
🛡 Enable `ServiceAccount` token projection and token requestor

rfranzke:

/area security
/kind enhancement
/merge squash

* When the extension is running on a seed with a gardenlet of at least `v1.37` then
  * `ServiceAccount` token projection is enabled.
  * the components deployed into shoot namespaces
    * uses `ServiceAccount` token projection.
    * uses a token managed by GRM's token requestor instead of a client certificate.
  * the Prometheus configuration is adapted. 

Part of gardener/gardener#4659
Part of gardener/gardener#4878

✅ ~Depends on #364, hence, PR is in draft state.~
✅ ~Depends on #381, hence, PR is in draft state.~

**Release note**:

```noteworthy operator
The extension controller uses a projected `ServiceAccount` token in case it runs on a seed with a gardenlet of at least `v1.37` or higher. Similarly, the components deployed into shoot namespaces will no longer use a client certificate but an auto-rotated `ServiceAccount` token which is only valid for `12h`.
```


dkistner (approved):
/lgtm

Issue #387, opened 2022-02-14
admission-gcp: Worker's `providerConfig.serviceAccount` is not properly validated


ialidzhikov:
**How to categorize this issue?**

/area quality
/kind bug
/platform gcp

**What happened**:
Currently admission-gcp allows the following WorkerConfig to be specified:
```yaml
          apiVersion: gcp.provider.extensions.gardener.cloud/v1alpha1
          kind: WorkerConfig
          serviceAccount:
            scopes:
              - 'https://www.googleapis.com/auth/cloud-platform'
```

Note that the `serviceAccount.name` should be required and non-empty. See https://github.com/gardener/gardener-extension-provider-gcp/blob/d22d912b74bab5f6665783cadfd41c112a351d64/pkg/apis/gcp/validation/worker.go#L53-L55

Checking the admission-gcp codebase, it looks like we have the following gap:

https://github.com/gardener/gardener-extension-provider-gcp/blob/d22d912b74bab5f6665783cadfd41c112a351d64/pkg/admission/validator/shoot.go#L140-L151

Note that `ValidateWorkerConfig` func (that also validates the `serviceAccount` field) is only invoked when there are `dataVolumes` specified.

**What you expected to happen**:
The `serviceAccount` field of the WorkerConfig should be validated properly and should not depend whether the pool has `dataVolumes` or not.

**How to reproduce it (as minimally and precisely as possible)**:

1. Specify the WorkerConfig from above

```yaml
          apiVersion: gcp.provider.extensions.gardener.cloud/v1alpha1
          kind: WorkerConfig
          serviceAccount:
            scopes:
              - 'https://www.googleapis.com/auth/cloud-platform'
```

2. Make sure this results in a MachineClass with the following spec

```yaml
    serviceAccounts:
    - email: ""
      scopes:
      - https://www.googleapis.com/auth/cloud-platform
```

3. machine-controller-manager validation fails. It cannot deal with such MachineClass:
```
[...]

PR #389, opened 2022-02-15
Validate Worker serviceAccount when workers dataVolumes are not provided

vlvasilev:

/area quality
/kind enhancement
/platform gcp

When worker.DataVolumes are not provided the [Worker ServiceAccount.Email check](https://github.com/gardener/gardener-extension-provider-gcp/blob/d22d912b74bab5f6665783cadfd41c112a351d64/pkg/admission/validator/shoot.go#L140-L151) is omitted. This PR prevents this from happening.
 
Fixes [#387](https://github.com/gardener/gardener-extension-provider-gcp/issues/387)

Thanks, @ialidzhikov for finding the issue.

**Release note**:

```bugfix user
The validation of the Worker.CloudProvider. ServiceAccount.Emai is no longer omitted and the field cannot be empty or invalid.
```

### DIFF
--- a/pkg/admission/validator/shoot.go
+++ b/pkg/admission/validator/shoot.go
@@ -140,13 +140,11 @@ func (s *shoot) validateContext(valContext *validationContext) field.ErrorList {
 	// WorkerConfig
 	for i, worker := range valContext.shoot.Spec.Provider.Workers {
 		workerFldPath := workersPath.Index(i)
-		for _, volume := range worker.DataVolumes {
-			workerConfig, err := admission.DecodeWorkerConfig(s.decoder, worker.ProviderConfig)
-			if err != nil {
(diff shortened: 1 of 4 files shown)


kon-angelo:
Maybe a small test case to prevent regression would be nice to have. Otherwise lgtm.

vlvasilev:
> Maybe a small test case to prevent regression would be nice to have. Otherwise lgtm.

Thanks, I have added an additional worker without `DataVolumes`. This should cover this case because a [test case](https://github.com/gardener/gardener-extension-provider-gcp/blob/17d771cb4d74aa23fa7ab9ddbc9e16b09a66bd2a/pkg/apis/gcp/validation/worker_test.go#L170) `should forbid because service account's email is empty` already exists.

kon-angelo (approved):
/lgtm

dkistner (approved):

PR #396, opened 2022-02-22
admission-gcp: Adapt Secrets webhook to rely on the provider label

ialidzhikov:
/area robustness
/area cost
/kind enhancement
/platform gcp

This PR adapts the Secrets webhook to rely on the provider type label (that is maintained by GCM). With this the Secrets webhook:
- no longer intercepts every Secret UPDATE request but only requests for Secrets with label `provider.shoot.gardener.cloud/gcp="true"`
- no longer needs to list Shoots (hence, no longer has a cache for Shoots based on Informers) to understand whether a Secret is in-use by a Shoot from the same provider type ("gcp")

Ref https://github.com/gardener/gardener/issues/4888

I think the component can be still improved to prevent get request (and cache) for SecretBindings - currently the Shoot validator on Shoot creation gets the SecretBinding -> the the Secret and validates the cloud provider Secret. This can be potentially moved to a new webhook for SecretBindings (when the `SecretBindingProviderValidator` feature gate is promoted to GA) and this get request (and cache) can be eliminated.

Part of #143

- [x] ✅   This PR relies on the provider type label to be maintained on the Secret, hence I will keep it in draft until the SecretBinding provider controller is enabled by default in GCM -> the controller is enabled by default for gardener-controller-manager >= 1.42 (ref https://github.com/gardener/gardener/pull/5499)

**Release note**:

```breaking operator
This version of admission-gcp requires the SecretBinding provider controller to be enabled - enabled by default for gardener-controller-manager >= 1.42 or can be enabled via the gardener-controller-manager component config.
```

```other operator
The Secrets webhook of admission-gcp:
- no longer intercepts every Secret UPDATE request but only requests for Secrets that are associated with a SecretBinding with `provider.type=gcp`.
- no longer needs to list Shoots (hence, no cache for Shoots)
```


timebertt:
Nice work @ialidzhikov!

PR #403, opened 2022-03-07
Allow projected volumes in PSP

timuthy:

/area security
/kind bug
/platform gcp

This PR fixes an issue with the CSI-Driver-Controller `DaemonSet` when `shoot.kubernetes.allowPrivilegedContainers: false` is set:

```
Error creating: pods "csi-driver-node-" is forbidden: unable to validate against any pod security policy: [spec.volumes[4]: Invalid value: "projected": projected volumes are not allowed to be used spec.
```

/cc @rfranzke

**Release note**:

```bugfix operator
An issue has been fixed with the `csi-driver-node` PodSecurityPolicy which blocked the creation of new CSI-Driver pods because `projected` volumes are not permitted.
```


ialidzhikov (approved):
/lgtm

kon-angelo (approved):
/lgtm

PR #411, opened 2022-03-16
Revendor gardener dependencies

kon-angelo:

/area control-plane
/kind enhancement
/platform gcp

Revendor gardener 

Fixes #

**Release note**:

```other operator
The following dependency is updated:
- github.com/gardener/gardener: v1.40.2 -> v1.42.3
```


kon-angelo:
cc @timebertt for `logcheck fix` commit

timebertt:
> cc @timebertt for `logcheck fix` commit

I'm sorry, but such comments are not really helpful.
For the next time: if you ask others for help, please give a detailed description about the issue you need help on.

timebertt:
I think, you are talking about this error message, right?:
```
$ make revendor
cannot find package "." in:
	/Users/timebertt/go/src/github.com/gardener/gardener-extension-provider-gcp/vendor/github.com/gardener/gardener/hack/tools
```

timebertt:
The error should be coming from https://github.com/gardener/gardener/blob/005aa840ac04e17e23e5ada6439632bbe138354b/hack/tools.mk#L25

The message, does not make anything fail – so that's good.
As long as you don't need `install-promtool.sh` or `logcheck` then the dependency is also not required.

Maybe we should simply enhance the line in g/g to not put out any errors on the missing (optional) dependency to minimize confusion?

acumino:
> The error should be coming from https://github.com/gardener/gardener/blob/005aa840ac04e17e23e5ada6439632bbe138354b/hack/tools.mk#L25
> 
> The message, does not make anything fail – so that's good. As long as you don't need `install-promtool.sh` or `logcheck` then the dependency is also not required.
> 
> Maybe we should simply enhance the line in g/g to not put out any errors on the missing (optional) dependency to minimize confusion?

That will be a better solution, should this PR be kept on hold, and vendor the next patch version of g/g.

timebertt:
> That will be a better solution, should this PR be kept on hold, and vendor the next patch version of g/g.

I mean, the error doesn't make anything fail, as said.
So you can either live with the message and go on with the PR or vendor a g/g version including https://github.com/gardener/gardener/pull/5588.

kon-angelo:
thank you @acumino @timebertt. If the message is just a warning I am okay with proceeding with the revendoring as-is or with a TODO item and we can improve later when we have a g/g release with your PR.

hack/tools.go
```diff
@@ -24,6 +24,10 @@ import (
 	_ "github.com/gardener/gardener/hack"
 	_ "github.com/gardener/gardener/hack/.ci"
 	_ "github.com/gardener/gardener/hack/api-reference/template"
+
+	// TODO(KA): This is imported to suppress the warning caused by the gardener/gardener tools. It can be
```
timebertt, line 28: You can simply drop 408d27a instead

kon-angelo: But then the warning will be printed. I think about it like so: if we already had to scratch our head then until we vendor the next gardener release, how many contributors will face the same ? I can live with an extra (unused) file in the repo instead.

go.mod
```diff
@@ -9,30 +9,30 @@ require (
...
+	golang.org/x/tools v0.1.9
 	google.golang.org/api v0.44.0
-	k8s.io/api v0.22.2
-	k8s.io/apiextensions-apiserver v0.22.2
-	k8s.io/apimachinery v0.22.2
-	k8s.io/apiserver v0.22.2
+	k8s.io/api v0.23.3
+	k8s.io/apiextensions-apiserver v0.23.3
+	k8s.io/apimachinery v0.23.3
+	k8s.io/apiserver v0.23.3
```
ialidzhikov, line 27: The replace directives have to be updated to `v0.23.3` as well.

rfranzke:
v1.42.3 is released with the fix

dkistner (approved):
/lgtm

ialidzhikov (approved):
Thanks!

/lgtm

PR #425, opened 2022-04-12
Forbid k8s version upgrade from `< v1.21` to `>= v1.21.0`

dkistner:
/kind enhancement
/platform gcp

As we are currently experience issues when upgrading kubernetes from `< v1.21` to `>= v1.21.0` as kcm is detaching all disks from the nodes during the process. [ref](https://github.com/kubernetes/kubernetes/issues/109354)

This is a safety measure to not affect existing cluster with the kcm detaching issue. We will remove this restriction once the issue is resolved. The upgrade can still be enforced by adding the annotation `shoot.gardener.cloud/force-version-upgrade=true` to the Shoot on own risk.

New Shoot can still be created with `>= v1.21.0`.

**Release note**:

```other user
GCP Shoot cluster can temporally not be upgrade from Kubernetes version < v1.21 to >= v1.21.0. The upgrade can still be enforced by adding the annotation `shoot.gardener.cloud/force-version-upgrade=true` to the Shoot on own risk.
```

### DIFF
--- a/pkg/admission/validator/shoot.go
+++ b/pkg/admission/validator/shoot.go
@@ -200,6 +200,9 @@ func (s *shoot) validateUpdate(ctx context.Context, oldShoot, currentShoot *core
 	allErrors = append(allErrors, gcpvalidation.ValidateWorkersUpdate(oldValContext.shoot.Spec.Provider.Workers, currentValContext.shoot.Spec.Provider.Workers, workersPath)...)
 	allErrors = append(allErrors, s.validateContext(currentValContext)...)
 
+	// TODO(dkistner) remove this once the kcm detaches all disks during upgrade from v1.20 to v1.21 is resolved.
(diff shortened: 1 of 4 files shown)


ialidzhikov:
I believe we can close this PR as there is now a recommended workaround for https://github.com/kubernetes/kubernetes/issues/109354:
1. Downgrade external-provisioner to v2.1.0 for K8s < 1.22 to prevent creating new PVs affected by the issue (incl. during the update from 1.20 to 1.21) -> already tackled by https://github.com/gardener/gardener-extension-provider-gcp/pull/439

2. Remove the `topology.kubernetes.io/zone` label from the affected PVs
   ```
   $ kubectl label pv -l topology.kubernetes.io/zone topology.kubernetes.io/zone-
   ```
 
 3. Perform the upgrade from 1.20 to 1.21
 
 /close

PR #438, opened 2022-05-13
Added GPU support for GCP machines

### LINKED ISSUE #132: Add support for Node Accelerators (GPU/TPU)
/component mcm
/kind enhancement
/priority normal
/platform gcp

In order to allow the usage of GPU/TPU nodes on GCP we need to enable the `guest_accelerators` when machines are being created.

In Terraform speak this would look something like this:

```shell
    guest_accelerator = [
      {
        type = "${var.gpu_type}"
        count = "${var.gpu_per_node}"
      }]
```

Of course this needs to be translated to the go sdk usage.

**Why is this needed**:
The use case here is to allow ML workload on GCP cluster.

himanshu-kun:

/area quality
/kind enhancement
/platform gcp

Added fields in `providerConfig` under worker section to allow the user to specify the gpu he would like to attach to his VM.
This configuration would be per worker pool. It looks like:
```
kind: WorkerConfig
gpu:
  acceleratorType: <type>
  count : <positive integer>
```

Minimal validation is added for these fields , and its left upto the provider API call to return relevant error in case gpu type is not supported or gpu type can't fit on particular machinetype.

**NOTE**: `onHostMigration` can't be set to `MIGRATE` in case VM has gpu attached and so its explicitly set to `TERMINATE` for gpu attached machines. 

*Also an update of the count or accelerator type would cause a rolling update for the worker pool*

Fixes #132 

* currently the following is left
  * add relevant documentation (Done)
  * test it locally to see fields are transported properly (Done)
  * test the effect of hostMaintainenceEvent on VM (Done)
  
**Release note**:

```feature user
Users could now attach GPU to their machines in a worker pool.
```


dhague (dismissed):

pkg/controller/worker/machines.go
```diff
@@ -75,7 +75,7 @@ func (w *workerDelegate) GenerateMachineDeployments(ctx context.Context) (worker
 	return w.machineDeployments, nil
 }
 
-func (w *workerDelegate) generateMachineConfig(ctx context.Context) error {
+func (w *workerDelegate) generateMachineConfig(_ context.Context) error {
```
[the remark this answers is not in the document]
himanshu-kun: looks like it was used earlier but later removed , and so not used now...will remove it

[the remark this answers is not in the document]
himanshu-kun: Just keeping it for future

pkg/apis/gcp/validation/worker.go
```diff
@@ -40,12 +40,31 @@ func ValidateWorkerConfig(workerConfig *gcp.WorkerConfig, dataVolumes []core.Dat
...
+
+	if gpu == nil {
+		return allErrs
+	}
+
+	if gpu.AcceleratorType == "" {
+		allErrs = append(allErrs, field.Required(fldPath.Child("acceleratorType"), "must be set when providing gpu"))
+	}
+
+	if gpu.Count == 0 {
```
unmarshall, line 61: to make the check more generic, can we change this to `gpu.Count <= 0` ?

pkg/controller/worker/machines.go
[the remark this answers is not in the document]
unmarshall: If this is required in the future then you have to anyways change the code to name this parameter. So instead of passing unused parameters to a method, you could also add this when it is required in the future. It is also possible that might not require it in which case it is just a technical debt to be cleaned up which can also be done now.

pkg/controller/worker/machines.go
```diff
@@ -224,8 +221,22 @@ func (w *workerDelegate) generateMachineConfig(ctx context.Context) error {
 					Region:       w.worker.Spec.Region,
 					Zone:         zone,
 				}
+
+				numGpus := pool.NodeTemplate.Capacity["gpu"]
```
unmarshall, line 225: In which cases would you additionally need to check  NodeTemplate.Capacity["gpu"] apart from `workerconfig.GPU` check which is done later in this method?

pkg/apis/gcp/validation/worker.go
himanshu-kun: okay

pkg/controller/worker/machines.go
himanshu-kun: none, just added a variable as it looked neater

unmarshall: ok this was a bit confusing because the value of `isLiveMigrationAllowed` is always going to be overwritten if `workerConfig.GPU` is not nil.

charts/internal/machineclass/values.yaml
```diff
@@ -23,6 +23,9 @@ machineClasses:
 #    image: projects/coreos-cloud/global/images/coreos-stable-1576-5-0-v20180105
 #    labels:
 #      name: my-disk
+  gpu:
+    acceleratorType: nvidia-tesla-t4
+    count: 1
```
kon-angelo, line 28: since the default is not to have any gpu, I would suggest you comment this section.

docs/usage-as-end-user.md
```diff
@@ -135,7 +135,16 @@ The worker configuration contains:
 
   Service accounts created in advance that generate access tokens that can be accessed through the metadata server and used to authenticate applications on the instance.
 
-An example `WorkerConfig` for the GCP looks as follows:
+* GPU with its type and count per node. This will attach that GPU to all the machines in the worker grp
+
+  **Note**: 
+  * A rolling upgrade of the worker group would be triggered in case the `acceleratorType` or `count` is updated.
+  * Some machineTypes like [a2 family](https://cloud.google.com/blog/products/compute/announcing-google-cloud-a2-vm-family-based-on-nvidia-a100-gpu) come with already attached gpu of `a100` type and pre-defined count. So DONOT specify workerConfig if workerPool is made of those machineTypes.
```
kon-angelo, line 142:
```suggestion
  * Some machineTypes like [a2 family](https://cloud.google.com/blog/products/compute/announcing-google-cloud-a2-vm-family-based-on-nvidia-a100-gpu) come with already attached gpu of `a100` type and pre-defined count. If your workerPool consists of those machineTypes, please **do not** specify any GPU configuration.
```

docs/usage-as-end-user.md
```diff
@@ -135,7 +135,16 @@ The worker configuration contains:
...
+  * Sufficient quota of gpu is needed in the GCP project. This includes quota to support autoscaling if enabled.
+  * GPU attached machines couldn't be live migrated during onHostMaintainence in GCP. Find out how to handle that in your application [here](https://cloud.google.com/compute/docs/gpus/gpu-host-maintenance)
```
kon-angelo, line 144:
```suggestion
  * GPU-attached machines can't be live migrated during host maintenance events. Find out how to handle that in your application [here](https://cloud.google.com/compute/docs/gpus/gpu-host-maintenance)
```

kon-angelo (approved):
/lgtm

kon-angelo (approved):
/lgtm

PR #450, opened 2022-06-06
Add MaxAllowed for VPA objects

dimitar-kostadinov:

/area auto-scaling
/kind enhancement
/platform gcp

Limits the range of maximum allowed resource requests to ensure that pods are schedulable
Fixes #

**Release note**:

```other operator

```


kon-angelo:
This is in draft for quite some time. I would like to ask you for the concrete reason behind this PR. When we recently removed the CPU limits we did it  because:
a) we saw a lot of throttling
b) these are critical system pods and getting them throttled could lead to system instability. Because of that we decided to had the upper limits removed.

cc: @voelzmo for extra opinion on the topic

charts/internal/seed-controlplane/charts/csi-driver-controller/values.yaml
```diff
@@ -73,4 +91,4 @@ csiSnapshotValidationWebhook:
       cpu: 10m
       memory: 32Mi
     limits:
-      memory: 200Mi 
+      memory: 200Mi
```
kon-angelo, line 94: Is it intentional that this is missing `maxAllowed` ?

kon-angelo (commented):
This PR only handles VPA of objects that we deploy in the seed ? Should we also do the same for the shoot system components e.g. the CSI-node drivers ?

dimitar-kostadinov:
>

charts/internal/seed-controlplane/charts/csi-driver-controller/values.yaml
dimitar-kostadinov: The [csi-driver-node](https://github.com/gardener/gardener-extension-provider-gcp/blob/8226169d598536b4384b13fcd7de5710caafb836/charts/internal/shoot-system-components/charts/csi-driver-node/templates/vpa.yaml#L5) on shoot is also marked as critical component. Thanks for pointing it out, I'll update the PR.

kon-angelo (approved):
/lgtm

kon-angelo (approved):
/lgtm

PR #451, opened 2022-06-07
Conditionally create serviceaccount and pass to mcm

kon-angelo:

/area control-plane
/kind enhancement
/platform gcp

GCP extension is currently always creating on infrastructure reconciliation a service account that is solely used for the VMs that MCM creates. By doing this unconditionally, it makes the GCP extension ton require ServiceAccount admin permissions. 

This PR changes the process in the following way:
a) if the user-provided SA has SA Admin permissions AND there is already a serviceaccount with the name of the shoot, provider-gcp will continue creating a service account and pass it to the machineclass
b) In any other case, don't create any SA.

In essence, [a] is preserving backwards-compatibility for existing shoots, but new shoots will be created without gardener's managed SA. 
Users can still opt to use their SA per worker pool. 

Fixes #

Depending on MCM change: https://github.com/gardener/machine-controller-manager-provider-gcp/pull/49

**Release note**:

```breaking users
Provider-GCP will no longer unconditionally create a service account to use for MCM machines. Existing shoots will continue to operate as-is, however no gardener-managed service account will be created for new shoots. Users creating new shoots are encouraged to use a new service account without the `ServiceAccount Admin` role for new shoots. 
```


pkg/gcp/serviceaccount.go
```diff
@@ -31,58 +31,66 @@ type ServiceAccount struct {
...
+// GetServiceAccountFromSecretReference retrieves the ServiceAccount from the secret with the given secret reference.
+func GetServiceAccountFromSecretReference(ctx context.Context, c client.Client, secretRef corev1.SecretReference) (*ServiceAccount, error) {
+	secret, err := extensionscontroller.GetSecretByReference(ctx, c, &secretRef)
 	if err != nil {
 		return nil, err
 	}
 
-	projectID, err := ExtractServiceAccountProjectID(data)
+	return GetServiceAccountFromSecret(secret)
+	// return GetServiceAccountFromJSON(data)
```
dkistner, line 46: remove line

pkg/gcp/serviceaccount.go
```diff
@@ -31,58 +31,66 @@ type ServiceAccount struct {
...
-	}, nil
+	return GetServiceAccountFromJSON(data)
 }
 
-// GetServiceAccountData retrieves the service account specified by the secret reference.
-func GetServiceAccountData(ctx context.Context, c client.Client, secretRef corev1.SecretReference) ([]byte, error) {
-	secret, err := extensionscontroller.GetSecretByReference(ctx, c, &secretRef)
-	if err != nil {
+// GetServiceAccountFromJSON returns a ServiceAccount from the given
+func GetServiceAccountFromJSON(data []byte) (*ServiceAccount, error) {
```
dkistner, line 59: Is there reuse outside of this package for this function?

pkg/controller/infrastructure/actuator_reconcile.go
```diff
@@ -59,9 +70,24 @@ func (a *actuator) reconcile(ctx context.Context, logger logr.Logger, infra *ext
...
+	createServiceAccount := true
+
+	_, err := iam.GetServiceAccount(context.Background(), clusterName)
+	if gcpclient.IgnoreErrorCodes(err, http.StatusNotFound, http.StatusUnauthorized) != nil {
+		return false, err
+	} else if err != nil {
+		createServiceAccount = false
+	}
+
+	return createServiceAccount, nil
```
dkistner, line 92:
```suggestion
func shouldCreateServiceAccount(iam gcpclient.IAMClient, clusterName string) (bool, error) {
	if _, err := iam.GetServiceAccount(context.Background(), clusterName); err != nil {
		if gcpclient.IsErrorCode(err, http.StatusNotFound, http.StatusUnauthorized) {
			return false, nil
		}
		return false, err
	}

	return true, nil
```

dkistner (commented):
Very nice pr. Thank you.
This is major step forward reducing required permissions that users need to provide.
I have only minor suggestions.

Once mcm-provider-gcp is able to cope with this as well and we integrated it we can go one from my point of view.

cc @donistz

dkistner (dismissed):
/lgtm

dkistner (approved):

PR #456, opened 2022-06-22
Change base image from alpine to distroless

dimityrmirchev:

/area open-source security
/kind enhancement
/platform gcp

This PR changes the base image for provider and admission services from `alpine` to [distroless](https://github.com/GoogleContainerTools/distroless). The processes will now use a non root user for their execution. This will reduce the attack surface of the images.

Fixes #

**Release note**:

```other operator
The extension now uses `distroless` instead of `alpine` as a base image.
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,6 +10,7 @@ FROM gcr.io/distroless/static-debian11:nonroot AS base
 
 ############# gardener-extension-provider-gcp
 FROM base AS gardener-extension-provider-gcp
+WORKDIR /
 
 COPY charts /charts
 COPY --from=builder /go/bin/gardener-extension-provider-gcp /gardener-extension-provider-gcp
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -6,17 +6,19 @@ COPY . .
 RUN make install
 
 ############# base image
-FROM alpine:3.13.7 AS base
+FROM gcr.io/distroless/static-debian11:nonroot AS base
```
ialidzhikov, line 9: In https://github.com/gardener/gardener/pull/6159 we changed the port of the gardener-apiserver from 443 to 8443
> so that it can be run as nonroot and avoid setting additional capabilities to that binary.

I see that admission-gcp is using 443 as well. Won't this break the admission-gcp image?

dimityrmirchev: Isn't [this](https://github.com/gardener/gardener-extension-provider-gcp/blob/e232602ca54048512a2c94c68ce5237a449d6741/charts/gardener-extension-admission-gcp/values.yaml#L26) the port the webhook is started on? The service port [here](https://github.com/gardener/gardener-extension-provider-gcp/blob/master/charts/gardener-extension-admission-gcp/charts/runtime/templates/service.yaml#L13) is OK to be left 443.

ialidzhikov: Okay, thanks for explaining!

ialidzhikov (approved):
/lgtm
/squash

kon-angelo (approved):
/lgtm

dkistner (approved):

PR #467, opened 2022-07-21
Adjusts VPA maxAllowed settings for MCM and CSI driver controller

dimitar-kostadinov:

/area auto-scaling
/kind enhancement
/platform gcp

Adjusts maxAllowed settings for MCM and CSI driver controller
Fixes #

**Release note**:

```other operator
NONE
```

### DIFF
--- a/charts/internal/machine-controller-manager/seed/templates/vpa.yaml
+++ b/charts/internal/machine-controller-manager/seed/templates/vpa.yaml
@@ -18,12 +18,15 @@ spec:
         cpu: {{ .Values.resources.mcmProviderGCP.requests.cpu }}
         memory: {{ .Values.resources.mcmProviderGCP.requests.memory }}
       maxAllowed:
-        cpu: {{.Values.resources.mcmProviderGCP.maxAllowed.cpu}}
-        memory: {{.Values.resources.mcmProviderGCP.maxAllowed.memory}}
+        cpu: {{ .Values.vpa.resourcePolicy.mcmProviderGCP.maxAllowed.cpu }}
+        memory: {{ .Values.vpa.resourcePolicy.mcmProviderGCP.maxAllowed.memory }}
       controlledValues: RequestsOnly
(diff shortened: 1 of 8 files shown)


charts/internal/machine-controller-manager/seed/templates/vpa.yaml
```diff
@@ -18,12 +18,15 @@ spec:
         cpu: {{ .Values.resources.mcmProviderGCP.requests.cpu }}
         memory: {{ .Values.resources.mcmProviderGCP.requests.memory }}
       maxAllowed:
-        cpu: {{.Values.resources.mcmProviderGCP.maxAllowed.cpu}}
-        memory: {{.Values.resources.mcmProviderGCP.maxAllowed.memory}}
+        cpu: 2
+        memory: 5G
```
kon-angelo, line 22: I am not sure I like this change very much. Why hardcode the values inside the templates and remove them from the values file ?

kon-angelo (changes requested):

charts/internal/machine-controller-manager/seed/templates/vpa.yaml
dimitar-kostadinov: I noticed that the `.Values.resources.*` is serialized in deployment template (e.g. [here](https://github.com/gardener/gardener-extension-provider-gcp/blob/003f0ac79ea5f8b03ea96b901168863ef8a38913/charts/internal/machine-controller-manager/seed/templates/deployment.yaml#L68)), but `maxAllowed` is not part of [resources](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#resources), so move it directly into the VPA templates

dimitar-kostadinov: Done

kon-angelo (approved):
/lgtm

PR #468, opened 2022-07-22
drop port 80 in firewall rule

### LINKED ISSUE #4: GCP: Control the default allow-external-access FW rules
We have Gardener shoot clusters running in GCP with firewall rules like the following.

```
shoot-prod-allow-external-access
Ingress
Apply to all    
IP ranges: 0.0.0.0/0
tcp:80,443
Allow
1000
```

The fact that the port 80 is open to the internet is being classified as a high security risk.
There is nothing from our services actually listening on that port and it will be good if we could close it. 

Is there a way for you to allow us to have control over that configuration and open it only if needed?

DockToFuture:

/area security
/kind enhancement
/platform gcp

Port 80 in external firewall rule is dropped.

Fixes #4

**Release note**:

```other operator
port 80 in external firewall rule is dropped.
```

### DIFF
--- a/pkg/internal/infrastructure/templates/main.tpl.tf
+++ b/pkg/internal/infrastructure/templates/main.tpl.tf
@@ -169,7 +169,7 @@ resource "google_compute_firewall" "rule-allow-external-access" {
 
   allow {
     protocol = "tcp"
-    ports    = ["80", "443"] // Allow ingress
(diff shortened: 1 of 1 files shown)


ScheererJ (approved):
/lgtm

kon-angelo (approved):
/lgtm

PR #471, opened 2022-07-25
Fix the Infrastructure integration test after #468

ialidzhikov:
/area testing
/kind bug
/platform gcp

**Release note**:

```other operator
NONE
```


kon-angelo (approved):
/lgtm

PR #473, opened 2022-07-26
allow infrastructure communication in pod range

DockToFuture:
**How to categorize this PR?**

/area networking
/kind enhancement
/platform gcp

**What this PR does / why we need it**:
To get rid of the networking overlay we need to allow the pod ip range communication within the vpc network.
Prerequisite for https://github.com/gardener/backlog/issues/29.

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Release note**:

```other operator
none
```

PR #474, opened 2022-07-28
run shoot cluster without overlay if not explicitly specified

DockToFuture:

/area networking
/kind enhancement
/platform gcp

A mutating webhook is added, so that all new calico gcp shoot clusters will be created without an overlay if not explicitly specified in the shoot spec.
Prerequisite for https://github.com/gardener/backlog/issues/29.

Requires https://github.com/gardener/gardener-extension-provider-gcp/pull/473 to work properly.

**Release note**:

```other operator
All new calico gcp shoot clusters will be created without an overlay if not explicitly specified in the shoot spec.
```


ScheererJ (approved):
/lgtm

pkg/webhook/network/mutate.go
```diff
@@ -0,0 +1,75 @@
...
+	if networkConfig.IPv4 == nil {
+		networkConfig.IPv4 = &ipv4
+	}
+
+	if networkConfig.IPv4 != nil && networkConfig.IPv4.Mode == nil {
+		networkConfig.IPv4.Mode = &ipv4PoolMode
+	}
+
+	if networkConfig.Backend == nil {
+		networkConfig.Backend = &backendNone
```
dkistner, line 67: Is setting the networking backend to `None` the important distinction between having an overlay or not (as I guess otherwise a backend like bird or vxlan would be used?)?

dkistner: cc @ScheererJ

ScheererJ: No, it is not. The `backend` option manages whether bird should be used for setting up local routes, e.g. `192.168.0.0/24 via 10.10.10.10`. This is not necessary if the routing is done by the network where kube/cloud-controller-manager manages the routes.
The important part to distinguish between overlay (ipip/vxlan) and no overlay is the block above, i.e. the ip mode. `Never` disables the overlay network.

dkistner: Thanks for clarification.

dkistner: So that means if a user provides a network config with another ip mode the overlay can still be used? It just disabled by default.

ScheererJ: Yes, we change the default, but the user can override at any point in time by specifying their own configuration.
This has the additional benefit that we can easily switch in case of issues.

dkistner: Nice. Thanks for the explanations.

PR #506, opened 2022-10-13
Blocking project wide ssh keys while creating machine class

rishabh-11:

/area control-plane
/kind enhancement
/platform gcp

In GCP, for each project, there is a `project-wide ssh key` which can be used to access all the VM instances in that project. Each Instance has an option to disable this by marking `Block project-wide SSH keys` as `On` from the GCP console. This PR adds a `metadata` field in the machine class spec during machine class creation, which will then be used by `mcm-provider-gcp` during VM instance creation to mark the `Block project wide SSH keys` field as `On`. The corresponding code in `mcm-provider-gcp` can be checked out [here](https://github.com/gardener/machine-controller-manager-provider-gcp/blob/ae925290549ce25b0116a5751ab3684a3d1a60b5/pkg/gcp/machine_controller_util.go#L129-L140). The `metadata` field will be present in the `providerSpec` of `machineclass` and will look like this:

Screenshot 2022-10-13 at 11 58 40 AM

Fixes #

The testing of this was done using a shooted seed local setup of gardener, and it worked as expected. The created machine had `Block project wide SSH keys` as `On` in the GCP console. 
Screenshot 2022-10-13 at 11 58 56 AM

**Release note**:

```feature user
`Block project wide SSH keys` will be `On` for all new machines created.
```

### DIFF
--- a/pkg/controller/worker/machines.go
+++ b/pkg/controller/worker/machines.go
@@ -179,7 +179,14 @@ func (w *workerDelegate) generateMachineConfig(_ context.Context) error {
 				"description":        fmt.Sprintf("Machine of Shoot %s created by machine-controller-manager.", w.worker.Name),
 				"disks":              disks,
 				"labels":             gceInstanceLabels,
-				"machineType":        pool.MachineType,
+				// TODO: make this configurable for the user
+				"metadata": []map[string]string{
+					{
+						"key":   "block-project-ssh-keys",
+						"value": "TRUE",
+					},
+				},
+				"machineType": pool.MachineType,
 				"networkInterfaces": []map[string]interface{}{
 					{
 						"subnetwork":        nodesSubnet.Name,
--- a/charts/internal/machineclass/values.yaml
+++ b/charts/internal/machineclass/values.yaml
@@ -28,6 +28,9 @@ machineClasses:
 #   count: 1
   labels:
     name: mcm
+  metadata:
+    - key: "block-project-ssh-keys"
+      value: "TRUE"
   machineType: n1-standard-4
   nodeTemplate:
     capacity:
--- a/pkg/controller/worker/machines_test.go
+++ b/pkg/controller/worker/machines_test.go
@@ -422,7 +422,13 @@ var _ = Describe("Machines", func() {
 								},
 							},
 						},
-						"labels":      gceInstanceLabels,
+						"labels": gceInstanceLabels,
+						"metadata": []map[string]string{
+							{
+								"key":   "block-project-ssh-keys",
+								"value": "TRUE",
+							},
+						},
 						"machineType": machineType,
 						"networkInterfaces": []map[string]interface{}{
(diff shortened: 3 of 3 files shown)


pkg/controller/worker/machines.go
```diff
@@ -179,7 +179,14 @@ func (w *workerDelegate) generateMachineConfig(_ context.Context) error {
 				"description":        fmt.Sprintf("Machine of Shoot %s created by machine-controller-manager.", w.worker.Name),
 				"disks":              disks,
 				"labels":             gceInstanceLabels,
-				"machineType":        pool.MachineType,
+				// TODO: make this configurable for the user
```
dkistner, line 182: Is there a concrete need already known to let users pass additional metadata for the machines?
If yes, we might wanna create an issue for it as it is better trackable?

rishabh-11: In the current `mcm-provider-gcp` code, we have a feature to add `metadata` field on a VM instance at the time of its creation(See [this](https://github.com/gardener/machine-controller-manager-provider-gcp/blob/ae925290549ce25b0116a5751ab3684a3d1a60b5/pkg/gcp/machine_controller_util.go#L129-L140)). But at the moment we cannot specify metadata field in the `providerConfig` section of the worker pool in the `shoot` yaml as the current `extension-provider-gcp` cannot decode it.(See [this](https://github.com/gardener/gardener-extension-provider-gcp/blob/3db08ee1f0c7ac279e47124360feeee841752675/pkg/apis/gcp/types_worker.go#L25) and [this](https://github.com/gardener/gardener-extension-provider-gcp/blob/3db08ee1f0c7ac279e47124360feeee841752675/pkg/controller/worker/machines.go#L137-L142)). So even though mcm provides the feature of specifying metadata on a VM instance, it will not be used as the machine class will never have this field present. This is the reason why I added this comment for now.

I am not opening an issue right now because we are unsure if we want to keep this field in the machine class. The reason is that any change in the `providerConfig` part of the `worker` section in the `shoot` yaml will generate a new machine class name (See [this](https://github.com/gardener/gardener-extension-provider-gcp/blob/3db08ee1f0c7ac279e47124360feeee841752675/pkg/controller/worker/machines.go#L206) and [this](https://github.com/gardener/gardener-extension-provider-gcp/blob/3db08ee1f0c7ac279e47124360feeee841752675/vendor/github.com/gardener/gardener/extensions/pkg/controller/worker/machines.go#L122)) and therefore a new machine class which will cause a rolling update of the machines. We don't want this just because some `metadata` field was changed. We already have an issue on mcm (See [this](https://github.com/gardener/machine-controller-manager/issues/750)) which requires something similar and should be sufficient to keep track of this as well.

cc @himanshu-kun

dkistner (approved):
/lgtm

kon-angelo (approved):
/lgtm

PR #511, opened 2022-10-21
Forbidding empty natIPNames array

### LINKED ISSUE #480: Empty natIPNames list for cloudNAT configuration should be rejected

/area networking
/kind bug
/platform gcp

**What happened**:

We had a user ticket where the shoot.yaml contained the following configuration for the `cloudNAT`:
```
networks:
        cloudNAT:
          minPortsPerVM: 2048
          natIPNames: []
```

This configuration was not rejected and caused an error during the Shoot creation:
```
task "Waiting until shoot infrastructure has been reconciled" failed: Error while waiting for Infrastructure shoot--hc-dev--i067095-haas/i067095-haas to become ready: error during reconciliation: Error reconciling infrastructure: could not find all requested variables: [nat_ips]
```

An empty list of `natIPNames` for `cloudNAT` should be rejected.

**Environment**:

- Extension version:
- Kubernetes version (use `kubectl version`): v1.22.11
- Others:

AleksandarSavchev:

/area networking
/kind bug
/platform gcp

This PR fixes the `networks` validation to forbid configuring an empty `natIPNames` array in `networks.cloudNAT`

Fixes #480

**Release note**:

```bugfix user
admission-gcp does no longer allow setting Shoot's `.spec.provider.infrastructureConfig.networks.cloudNAT.natIPNames` field to empty list.
```

### DIFF
--- a/pkg/apis/gcp/validation/infrastructure.go
+++ b/pkg/apis/gcp/validation/infrastructure.go
@@ -124,8 +124,8 @@ func ValidateInfrastructureConfig(infra *apisgcp.InfrastructureConfig, nodesCIDR
 		}
 	}
 
-	if infra.Networks.CloudNAT != nil && len(infra.Networks.CloudNAT.NatIPNames) == 0 {
-		allErrs = append(allErrs, field.Forbidden(networksPath.Child("cloudNAT", "natIPNames"), "should not be empty array"))
+	if infra.Networks.CloudNAT != nil && infra.Networks.CloudNAT.NatIPNames != nil && len(infra.Networks.CloudNAT.NatIPNames) == 0 {
+		allErrs = append(allErrs, field.Invalid(networksPath.Child("cloudNAT", "natIPNames"), infra.Networks.CloudNAT.NatIPNames, "nat IP names cannot be empty"))
 	}
 
 	return allErrs
--- a/pkg/apis/gcp/validation/infrastructure.go
+++ b/pkg/apis/gcp/validation/infrastructure.go
@@ -124,6 +124,10 @@ func ValidateInfrastructureConfig(infra *apisgcp.InfrastructureConfig, nodesCIDR
 		}
 	}
 
+	if infra.Networks.CloudNAT != nil && len(infra.Networks.CloudNAT.NatIPNames) == 0 {
+		allErrs = append(allErrs, field.Forbidden(networksPath.Child("cloudNAT", "natIPNames"), "should not be empty array"))
+	}
+
 	return allErrs
 }
 
--- a/pkg/apis/gcp/validation/infrastructure_test.go
+++ b/pkg/apis/gcp/validation/infrastructure_test.go
@@ -312,13 +312,20 @@ var _ = Describe("InfrastructureConfig validation", func() {
 				errorList := ValidateInfrastructureConfig(newInfrastructureConfig, &nodes, &pods, &services, fldPath)
 				Expect(errorList).To(BeEmpty())
 			})
-			It("should allow correct CloudNAT config", func() {
+			It("should allow CloudNAT config with valid NatIPNames", func() {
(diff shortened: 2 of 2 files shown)


pkg/controller/infrastructure/configvalidator.go
```diff
@@ -76,7 +76,12 @@ func (c *configValidator) Validate(ctx context.Context, infra *extensionsv1alpha
 func (c *configValidator) validateNetworks(ctx context.Context, computeClient gcpclient.ComputeClient, clusterName, region string, networks api.NetworkConfig, fldPath *field.Path) field.ErrorList {
 	allErrs := field.ErrorList{}
 
-	if networks.CloudNAT == nil || len(networks.CloudNAT.NatIPNames) == 0 {
+	if networks.CloudNAT == nil {
+		return allErrs
+	}
+
+	if len(networks.CloudNAT.NatIPNames) == 0 {
```
ialidzhikov, line 83: The validation should be rather part of `admission-gcp` - https://github.com/gardener/gardener-extension-provider-gcp/blob/37e1e2ca92132db555271ed286f8d3d337253085/pkg/apis/gcp/validation/infrastructure.go#L28 . 

This is to prevent created/updating Shoots with empty `networks.cloudNAT.natIPNames`.
The `ConfigValidator` validation is executed before Infrastructure reconciliation when the Shoot is already created/updated with the invalid spec. And the `ConfigValidator` was mainly introduced to allow validating resources using the cloud provider APIs as well. You can read more about the concept in https://github.com/gardener/gardener/blob/v1.58.0/docs/extensions/infrastructure.md#configvalidator-interface.

ialidzhikov (changes requested):

pkg/apis/gcp/validation/infrastructure.go
```diff
@@ -124,6 +124,10 @@ func ValidateInfrastructureConfig(infra *apisgcp.InfrastructureConfig, nodesCIDR
 		}
 	}
 
+	if infra.Networks.CloudNAT != nil && len(infra.Networks.CloudNAT.NatIPNames) == 0 {
+		allErrs = append(allErrs, field.Forbidden(networksPath.Child("cloudNAT", "natIPNames"), "should not be empty array"))
```
ialidzhikov, line 128: Nit: we could use `field.Invalid` instead of `field.Forbidden`:

```suggestion
		allErrs = append(allErrs, field.Invalid(networksPath.Child("cloudNAT", "natIPNames"), infra.Networks.CloudNAT.NatIPNames, "nat IP names cannot be empty"))
```

pkg/apis/gcp/validation/infrastructure.go
```diff
@@ -124,6 +124,10 @@ func ValidateInfrastructureConfig(infra *apisgcp.InfrastructureConfig, nodesCIDR
 		}
 	}
 
+	if infra.Networks.CloudNAT != nil && len(infra.Networks.CloudNAT.NatIPNames) == 0 {
```
ialidzhikov, line 127: Correct me if I am wrong but this validation will also start failing for valid InfrastructureConfig such as:
```yaml
networks:
  cloudNAT:
    minPortsPerVM: 2048
```

We should rather require at least 1 nat IP names to be provided **only** when the field is explicitly set to empty array.
We should not require at least 1 nat IP name to be provided **always** because the field is optional.

The check we need is:
```suggestion
	if infra.Networks.CloudNAT != nil && infra.Networks.CloudNAT.NatIPNames != nil && len(infra.Networks.CloudNAT.NatIPNames) == 0 {
```

because `len(nil)` is `0`.

Can you also add an unit test to cover this scenario?

ialidzhikov (changes requested):

ialidzhikov (approved):
/lgtm
/squash

PR #518, opened 2022-11-08
Enabled uniform access level for bucket creation in GCP.

abdasgupta:

/area TODO
/kind TODO
/platform gcp

This PR enables the creation of bucket with uniform access levels in GCP.
Fixes #
#508

I am not too familiar with the codebase in this repo. I couldn't add any test case for the changes I made. Are the test cases necessary for my changes? If yes, how to add them?

**Release note**:

```feature operator
GCP buckets are created with Uniform access level.
```

### DIFF
--- a/pkg/gcp/client/storage.go
+++ b/pkg/gcp/client/storage.go
@@ -70,6 +70,9 @@ func (s *storageClient) CreateBucketIfNotExists(ctx context.Context, bucketName,
 	if err := s.client.Bucket(bucketName).Create(ctx, s.serviceAccount.ProjectID, &storage.BucketAttrs{
 		Name:     bucketName,
 		Location: region,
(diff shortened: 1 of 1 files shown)


kon-angelo:
hello @abdasgupta. Could you please clarify the following points:
- The code update is only for new buckets getting created. What about existing buckets ? Can we do something similar ?
- Do we need to make updates to our existing IAM permissions ?

abdasgupta:
Hi @kon-angelo Yes, the code update is only for new buckets created. For existing buckets, we have to change the access level manually ( or maybe, write a tool that will do this automatically on a human's behalf). I have already changed access levels for those buckets for which alerts were raised.

I didn't get your second question. I don't think IAM permissions are involved here, but I am not sure

kon-angelo (approved):
/lgtm

PR #522, opened 2022-11-21
disable network overlay for cilium by default and add tests

DockToFuture:

/area networking
/kind enhancement
/platform gcp

Overlay networks for newly created cilium shoot clusters are dropped now by default. To keep the overlay network `networking.providerConfig.overlay.enabled: true` needs to be set explicitly in the shoot spec.
Additional tests were added to the shoot mutation.
 
Fixes #

**Release note**:

```other operator
Overlay networks for newly created cilium shoot clusters are dropped now by default. To keep the overlay network `networking.providerConfig.overlay.enabled: true` needs to be set explicitly in the shoot spec.
```


kon-angelo (approved):
/lgtm

PR #539, opened 2022-12-14
Check for existing GCP service account using TF state

MartinWeindel:

/area control-plane
/kind enhancement
/platform gcp

The decision if a GCP service account needs to be created is changed to rely on the terraformer state.
If there is already a service account, it is kept.
For new clusters one is only created if the feature gate `DisableGardenerServiceAccountCreation` is set to `false`. 

Fixes #

**Release note**:

```other operator
Check for existing GCP service account using TF state
```

### DIFF
--- a/pkg/controller/infrastructure/actuator_reconcile.go
+++ b/pkg/controller/infrastructure/actuator_reconcile.go
@@ -17,16 +17,16 @@ package infrastructure
 import (
 	"context"
 	"fmt"
-	"net/http"
 
 	"github.com/gardener/gardener/extensions/pkg/controller"
 	"github.com/gardener/gardener/extensions/pkg/terraformer"
 	extensionsv1alpha1 "github.com/gardener/gardener/pkg/apis/extensions/v1alpha1"
 	"github.com/go-logr/logr"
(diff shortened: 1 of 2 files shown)


pkg/controller/infrastructure/actuator_reconcile.go
```diff
@@ -17,18 +17,19 @@ package infrastructure
...
 	"github.com/go-logr/logr"
 
 	"github.com/gardener/gardener-extension-provider-gcp/pkg/apis/gcp/helper"
+	"github.com/gardener/gardener-extension-provider-gcp/pkg/controller/infrastructure/infraflow/shared"
 	"github.com/gardener/gardener-extension-provider-gcp/pkg/features"
-	gcpclient "github.com/gardener/gardener-extension-provider-gcp/pkg/gcp/client"
 	"github.com/gardener/gardener-extension-provider-gcp/pkg/internal"
 	"github.com/gardener/gardener-extension-provider-gcp/pkg/internal/infrastructure"
+
+	"k8s.io/apimachinery/pkg/runtime"
```
kon-angelo, line 32: I think this needs reformating

kon-angelo (approved):
/lgtm

Issue #558, opened 2023-02-21
Only configure pod routes in VPC if pod overlay network is disabled


ScheererJ:
**How to categorize this issue?**

/area networking
/area control-plane
/kind enhancement
/platform gcp

**What would you like to be added**:
If overlay network is enabled the cloud routes should not be enabled.

**Why is this needed**:
https://cloud.google.com/vpc/docs/quota#vpc-peering describes that GCP has a default limit of 300 static routes. If a cluster grows beyond 300 nodes or the VPC is shared among multiple clusters with together over 300 nodes the limit might be hit even though the cloud routes are not required if overlay network is enabled.

kon-angelo:
Resolved with: https://github.com/gardener/gardener-extension-provider-gcp/pull/561

PR #561, opened 2023-02-27
optionally configure cloud routes based on overlay

kon-angelo:

/area control-plane
/kind bug
/platform gcp

Do not configure cloud routes if you are not using overlay network

Fixes #https://github.com/gardener/gardener-extension-provider-gcp/issues/558

**Release note**:

```other operator
Enable cloud-controller-manager's route controller only if the shoot is using an overlay network. 
```

### DIFF
--- a/pkg/controller/controlplane/valuesprovider.go
+++ b/pkg/controller/controlplane/valuesprovider.go
@@ -20,10 +20,15 @@ import (
 	"path/filepath"
 	"strings"
 
+	calicov1alpha1 "github.com/gardener/gardener-extension-networking-calico/pkg/apis/calico/v1alpha1"
+	"github.com/gardener/gardener-extension-networking-calico/pkg/calico"
+	ciliumv1alpha1 "github.com/gardener/gardener-extension-networking-cilium/pkg/apis/cilium/v1alpha1"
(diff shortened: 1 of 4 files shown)


pkg/controller/controlplane/valuesprovider.go
```diff
@@ -503,3 +514,40 @@ func getNetworkNames(
...
+			return false, err
+		}
+		o := networkConfig.Overlay
+		if o == nil {
+			return true, nil
+		}
+		return o == nil || o.Enabled, nil
+	}
+
+	return true, nil
```
ScheererJ, line 552: Would it not be a good idea to return an error here? This is certainly not an expected case in that we do not know the networking type.

ScheererJ (commented):
Please invert the release note.

pkg/controller/controlplane/valuesprovider.go
kon-angelo: Not sure. That would strictly limit the networking extensions to the very limited set of {calico,cilium} and hardcode that dependency on the infra provider.

ScheererJ: Sure, your call.

ScheererJ (approved):
/lgtm

MartinWeindel (approved):
/lgtm

PR #571, opened 2023-03-17
add support for eim

kon-angelo:

/area control-plane
/kind enhancement
/platform gcp

GCP by default uses endpoint-dependent mapping. However we see that our current TF scripts default to enabling the Endpoint-Independent Mapping instead even though we don't explicitly specify that. This causes issues for users due to endpoint_independence_conflicts that occur when making multiple connections to the kube-apiserver in large clusters. 

This PR changes the default behavior to endpoint-dependent mapping and exposes a new option that users can use to customize the behavior they want. 

cc: @ScheererJ 

Fixes #

**Release note**:

```breaking user
The default NAT mapping has changed to endpoint-dependent. Existing connections should not be interrupted but users will need to adapt their configuration if they want to keep Endpoint-Independent Mapping for their CloudNAT instance.
```
```feature user
Users can now specify whether they want to enable or disable Endpoint-Independent Mapping.
```


MartinWeindel (approved):
/lgtm

MartinWeindel (commented):
The field `endpointIndependentMapping` should also be shown in the example InfrastructureConfig (line 61 of this file)

MartinWeindel (approved):
/lgtm

docs/usage-as-end-user.md
```diff
@@ -89,6 +91,8 @@ The `networks.cloudNAT.minPortsPerVM` is optional and is used to define the [min
 
 The `networks.cloudNAT.natIPNames` is optional and is used to specify the names of the manual ip addresses which should be used by the nat gateway
 
+The `networks.cloudNAT.endpointIndependentMapping` is optional and is used to define the [endpoint mapping behavior](https://cloud.google.com/nat/docs/ports-and-addresses#ports-reuse-endpoints).
```
ScheererJ, line 94: Do you want to specify in the documentation what is the default (on/off)?

ScheererJ (commented):
Looks good. I would prefer to document the default value as it is not obvious what happens when you do not specify the new option.

PR #591, opened 2023-05-04
Setting CSI GCP topology label on machine deployment creation to allo…

elankath:
…w scale-up

/area auto-scaling
/kind enhancement
/platform gcp

adds driver specific CSI label `topology.gke.io/zone` during machineDeployment creation

Fixes #430

**Release note**:

```other operator
machineDeployment will have the label `topology.gke.io/zone` when created.
```


pkg/controller/worker/machines.go
```diff
@@ -88,6 +92,9 @@ func (w *workerDelegate) generateMachineConfig(_ context.Context) error {
 		machineClasses     []map[string]interface{}
 		machineImages      []apisgcp.MachineImage
 	)
+	const (
+		csiMigrationVersion = "1.21"
```
kon-angelo, line 96: Gardener has already dropped support for anything bellow 1.20 so the check is not necessary. Regardless the CSI migration version for gcp was 1.18 if I am not mistaken.

pkg/controller/worker/machines.go
```diff
@@ -44,6 +46,8 @@ const (
 	maxGcpLabelCharactersSize = 63
 	// ResourceGPU is the GPU resource . It should be a non-negative integer.
 	ResourceGPU v1.ResourceName = "gpu"
+	// CSIDiskDriverTopologyKey is the label on persistent volumes that represents availability by zone.
+	CSIDiskDriverTopologyKey = "topology.gke.io/zone"
```
kon-angelo, line 50: Out of convention, I would probably move this key to https://github.com/gardener/gardener-extension-provider-gcp/blob/master/pkg/gcp/types.go

kon-angelo (changes requested):
Thank you for the contribution. Small comments from my side

pkg/controller/worker/machines.go
elankath: Ok, was un-certain about this. Will remove the check for CSI. thanks.

pkg/controller/worker/machines.go
```diff
@@ -168,6 +170,9 @@ func (w *workerDelegate) generateMachineConfig(_ context.Context) error {
 		gceInstanceLabels := getGceInstanceLabels(w.worker.Name, pool)
 		isLiveMigrationAllowed := true
 
+		if err != nil {
+			return err
+		}
```
kon-angelo, line 175: Is this still necessary ?

kon-angelo (approved):
/lgtm

elankath:
thanks Konstantinos.

PR #597, opened 2023-05-15
topology-aware routing: Adapt to the upstream annotation change in K8s 1.27

ialidzhikov:

/area high-availability
/area networking
/kind enhancement
/platform gcp

See https://github.com/gardener/gardener/pull/7933

N/A

N/A

**Release note**:

```other operator
NONE
```

### DIFF
--- a/charts/gardener-extension-admission-gcp/charts/runtime/templates/service.yaml
+++ b/charts/gardener-extension-admission-gcp/charts/runtime/templates/service.yaml
@@ -5,9 +5,13 @@ metadata:
   namespace: {{ .Release.Namespace }}
   annotations:
     networking.resources.gardener.cloud/from-all-webhook-targets-allowed-ports: '[{"protocol":"TCP","port":{{ .Values.global.webhookConfig.serverPort }}}]'
-  {{- if .Values.global.service.topologyAwareRouting.enabled }}
+    {{- if .Values.global.service.topologyAwareRouting.enabled }}
+    {{- if semverCompare ">= 1.27-0" .Capabilities.KubeVersion.GitVersion }}
+    service.kubernetes.io/topology-mode: "auto"
+    {{- else }}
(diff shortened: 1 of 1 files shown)


charts/gardener-extension-admission-gcp/charts/runtime/templates/service.yaml
```diff
@@ -5,9 +5,13 @@ metadata:
   namespace: {{ .Release.Namespace }}
   annotations:
     networking.resources.gardener.cloud/from-all-webhook-targets-allowed-ports: '[{"protocol":"TCP","port":{{ .Values.global.webhookConfig.serverPort }}}]'
-  {{- if .Values.global.service.topologyAwareRouting.enabled }}
+    {{- if .Values.global.service.topologyAwareRouting.enabled }}
+    {{- if semverCompare ">= 1.27-0" .Capabilities.KubeVersion.GitVersion }}
```
kon-angelo, line 9: Just a question: Am I correct that the `-0` a prerelease check and do we need it ?

ialidzhikov: Initially I copied this statement from gardener/gardener. After having a deeper look now:

I think `-0` makes possible to ignore suffixes. See https://github.com/gardener/gardener/pull/7207. That's the PR add added the `-0` suffixes in the version pkg in gardener:
> This PR adapts the Golang-based version constraint matchers to ignore suffixes (this is also what we do in the Helm-based version constraints).
If you search for `emverCompare ">=` in gardener/gardener, all versions are with the `-0` suffix.

We need to be able to ignore suffixes especially for admission-gcp, as in our setup we use GKE and the K8s versions there have a suffix (for example `v1.x.x-gke.3200`).

kon-angelo (approved):
/lgtm

PR #607, opened 2023-05-31
Support for CMEK disk encryption of vols

### LINKED ISSUE #564: Enable CMEK Encrypted Shoot Worker Node Boot Disks

/area security
/kind enhancement

We need the ability to deploy Gardener Shoot Clusters to GCP where the shoot worker node root volumes (node boot disk) are encrypted using Customer Managed Encryption Keys (CMEK).  It appears that we can configure additional PVCs to use CMEK by configuring an appropriate storage class.  This is our primary feature request.  

Additionally, once shoot clusters can be configured to be deployed with CMEK on the node boot disks, we'd like to also be able to deploy managed seeds that are also deployed with CMEK encrypted node boot disks.  

**Why is this needed**:
Customer requirements.

elankath:

/area security
/kind enhancement
/platform gcp

Support for CMEK disk encryption for volumes. 
Fixes #564 
Related: https://github.com/gardener/machine-controller-manager-provider-gcp/issues/78

Request review and suggestions for extra tests. 
**Release note**:

```other operator
Support for CMEK Disk Encryption for volumes
```


himanshu-kun:
The PR is merged @kon-angelo

himanshu-kun (commented):
I see that currently you have encrypted all the disks (root + data) if cmek key is provided , irrespective of checking `volume.encrypted` value.

I did some more research on this and found the following:
* in AWS there is an option on each disk to encrypt and decrypt and so the extension-aws sets it to [true by default](https://github.com/gardener/gardener-extension-provider-aws/blob/96587fba6fb2f58b0943bd015ddf070e8d749a5b/pkg/controller/worker/machines.go#L314)
* in GCP disks (compute.AttachedDisk) there is no such option and the docstring of `compute.AttachedDisk.DiskEncryptionKey` field suggests that the disks are encrypted by default encryption everytime (Note: there is no option to keep the disks un-encrypted like AWS)
* because of above reason , gcp-extension [**doesn't use `volume.Encryption`**](https://github.com/gardener/gardener-extension-provider-gcp/blob/502fc6eadf8c36d7ebdb6254b1e5676250bef831/pkg/controller/worker/machines.go#L283)

This means user providing `volume.Enryption`=`true/false` doesn't affect anything about the disk encryption in case of GCP disks even now.

So with CMEK encryption in , I think we should keep it the same, but also update the documentation about the irrelevance of setting `volume.Encryption` in GCP worker resource
@kon-angelo WDYT?

pkg/controller/worker/machines.go
```diff
@@ -151,6 +154,12 @@ func (w *workerDelegate) generateMachineConfig(_ context.Context) error {
 				disk["interface"] = *workerConfig.Volume.LocalSSDInterface
 			}
 
+			if workerConfig.Volume != nil && *volume.Type != "SCRATCH" {
```
kon-angelo, line 157: I think you should also check for `volume.Type != nil` here

kon-angelo:
Regarding documentation:

+ You should add a (commented) example: in https://github.com/gardener/gardener-extension-provider-gcp/blob/274c4b9c1e04ac9b8a328bf4be70badb274ca5d8/example/30-worker.yaml#L88
+ Add some small explanation https://github.com/gardener/gardener-extension-provider-gcp/blob/master/docs/usage-as-end-user.md#workerconfig.

charts/internal/machineclass/values.yaml
```diff
@@ -15,6 +15,11 @@ machineClasses:
     image: projects/coreos-cloud/global/images/coreos-stable-1576-5-0-v20180105
     labels:
       name: my-disk
+#   CMEK Encryption
+#     encryption:
+#       kmsKeyName: projects/projId/locations/zone/keyRings/keyRingName/cryptoKeys/keyName
+#       kmsKeyServiceAccount: name@projId.iam.gserviceaccount.com
```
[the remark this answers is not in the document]
elankath: fixed

pkg/controller/worker/machines.go
elankath: yep thats right. fixed.

kon-angelo (approved):
/lgtm

kon-angelo (approved):
/lgtm

PR #608, opened 2023-06-01
Update csi-driver to v1.9.5

vpnachev:
**How to categorize this PR?**

/platform gcp

**What this PR does / why we need it**:
Update csi-driver to v1.9.5

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:
The new image should fix an issue with XFS filesystems, see https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/1204

**Release note**:

```other operator
The GCP csi-driver has been updated to v1.9.5. 
```

PR #611, opened 2023-06-06
Revert "Update csi-driver to v1.9.5"

vpnachev:
Reverts gardener/gardener-extension-provider-gcp#608

The OCI image is still not available (see https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/issues/1250), once https://github.com/kubernetes/k8s.io/pull/5360 is merged it will be promoted from the staging registry and we can upgrade again the version.


kon-angelo (approved):
/lgtm

PR #613, opened 2023-06-07
always configure cloud-routes

kon-angelo:

/area control-plane
/kind regression
/platform gcp

There is a race condition between the CCM's node controller and calico when the route controller is disabled. Concretely the node controller sets the `NetworkUnavailable` to true in all cases. On the other hand route controller unsets the condition when it creates the route. If the node controller reconciles a new node after calico, then a node healthy will be marked with the above condition. 
Therefore while the problem persists, configure routes regardless of if overlay isused.

Fixes #

**Release note**:

```other operator
Update CCM configuration to always enable the route controller regardless if overlay is used. This is done to prevent a race condition that would mark an otherwise healthy node with the `NetworkUnavailable` condition.
```

### DIFF
--- a/pkg/controller/controlplane/valuesprovider.go
+++ b/pkg/controller/controlplane/valuesprovider.go
@@ -428,11 +428,10 @@ func (vp *valuesProvider) getCCMChartValues(
 		values["featureGates"] = cpConfig.CloudControllerManager.FeatureGates
 	}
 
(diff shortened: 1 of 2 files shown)


MartinWeindel (approved):
/lgtm

dkistner (approved):
/lgtm

PR #631, opened 2023-07-12
Mitigate node condition NetworkUnavailable being set by node controller in a race with CNI

ScheererJ:

/area networking
/area robustness
/area scalability
/kind enhancement
/platform gcp

Mitigate node condition NetworkUnavailable being set by node controller in a race with CNI.

A new mutating webhook is now always added to shoot clusters managed by provider gcp.
It will override setting node condition NetworkUnavailable by cloud controller manager's
node controller. This resolves the race between node controller, route controller and CNI.
Furthermore, it allows cloud controller manager's route controller to be disabled again
for clusters using overlay network.

Enable cloud controller manager's route controller only when needed.

This change reverts https://github.com/gardener/gardener-extension-provider-gcp/pull/613. With the mitigation for node condition
NetworkUnavailable in place, it is possible to only activate route
controller when required and disable it if overlay network is used.
This allows clusters to grow well beyond GCP's route limits if they
use overlay network.

None.

**Release note**:

```other operator
`cloud-controller-manager`'s route controller is no longer activated for clusters with overlay network
```


timuthy (commented):
/lgtm

kon-angelo:
Just a small note that I want to document in case we need it for future use:

You can technically narrow the conditions for the mutation even further for the node controller by checking if the node object also had the cloud-provider's `unintialized` taint. 

However as of now there would be no difference in using it.

kon-angelo (approved):
/lgtm

PR #637, opened 2023-08-08
Update region config example

timuthy:

/area scalability
/kind enhancement
/platform gcp

This PR updates the region config example.

**Release note**:

```other operator
NONE
```


kon-angelo:
Minor nit: Can we keep the alphabetical order ? It see that now it is messed up and I can imagine that it will be easier to reason about changes with it.

timuthy:
@kon-angelo this was actually done on purpose (BTW why do you think it's messed up? 😄): If you search for a specific region in the config, then you see the list ordered by distance which helps to better compare the values.

kon-angelo:
@timuthy Ah, my bad. I was focused on the "key" part of the key-value. Then let's proceed

PR #651, opened 2023-10-06
Remove dependency to specific calico and cilium versions

axel7born:

/area networking
/kind technical-debt
/platform gcp

This PR removes the dependency to specific calico and cilium versions.
After each api change of a network extension, the new version needed to be vendored here and a new release has to be made in order to pass admission.
By parsing the network into a general map instead into an object of type `NetworkConfig` we can get rid of this dependency. 

Fixes #

**Release note**:

```noteworthy developer
Remove dependency to specific calico and cilium versions.
```


kon-angelo (approved):
/lgtm

PR #664, opened 2023-11-10
Add new mutator tests to clarify expected behavior.

axel7born:

/area networking
/kind test
/platform gcp

Add tow more test:
* Check that arbitrary network config is not lost.
* Check that existing shoots with no overlay config don't get one.
Fixes #

**Release note**:

```other developer
Add new unit tests.
```


kon-angelo (approved):
/lgtm

PR #660, opened 2023-11-07
Add a label to the disks to associate them with a Shoot cluster

ialidzhikov:
**How to categorize this PR?**

/area storage
/kind enhancement
/platform gcp

**What this PR does / why we need it**:
Right now it is impossible to associate a disk created for a Shoot cluster with the Shoot cluster itself. The disk name is in format `pv--<uid>` (e.g. `pv--f7bcfb30-5833-4b38-a530-fdfd844fcba1`). The disk itself does not have any labels. Hence, there is nothing that could hint what is the Shoot cluster owning this disk.
Having such a label would fix the above-described problem and would potentially help a machinery that checks for orphan resources.

---

I collected how the volume/disk tags/labels changed over time for all cloud providers:

- AWS volumes created with the in-tree provisionner (`kubernetes.io/aws-ebs`):
Tags:
```
KubernetesCluster                        shoot--foo--bar1     
Name                                     shoot--foo--bar1-dynamic-pvc-a78c281f-78d5-46cd-8849-3fc2bd04ab28     
kubernetes.io/cluster/shoot--foo--bar1   owned     
kubernetes.io/created-for/pv/name        pvc-a78c281f-78d5-46cd-8849-3fc2bd04ab28     
kubernetes.io/created-for/pvc/name       pvc-name
kubernetes.io/created-for/pvc/namespace  default
```

- AWS volumes created with the out-of-tree provisionner (`ebs.csi.aws.com`):
Tags:
```
CSIVolumeName                             pv-shoot--foo--bar2-40e919e4-95f2-4808-8539-a4d0547b4aff     
Name                                      shoot--foo--bar2-dynamic-pv-shoot--foo--bar2-40e919e4-95f2-4808-8539-a4d0547b4aff
kubernetes.io/cluster/shoot--foo--bar2    owned
kubernetes.io/created-for/pv/name         pv-shoot--foo--bar2-40e919e4-95f2-4808-8539-a4d0547b4aff
kubernetes.io/created-for/pvc/name        pvc-name
kubernetes.io/created-for/pvc/namespace   default
```

Resolved with https://github.com/gardener/gardener-extension-provider-aws/pull/256

---

- GCP volumes created with the in-tree provisionner (`kubernetes.io/gce-pd`):
Volume name format: `<shoot-techincal-id>-dynam-pvc-603cec13-8904-492b-9b21-980ee2bb8411`

Labels: No labels but the description contains as string:
```
[...]

PR #669, opened 2023-12-11
Enhance labels for gcp workers and disks

AndreasBurger:
/kind enhancement
/area usability
/platform gcp

With this PR the labeling is made more consistent for GCP-Workers and -Disks:
- Disks attached to instances on creation will also receive the same labels as the instance itself.
- An additional label `k8s-cluster-name` is attached to instances (and thus boot disks). This label is already in place for disks created via PV since #660.

Together this should improve the user-experience by providing a consistent label across instances and _all_ disks, `k8s-cluster-name`. Also, user-added labels are now propagated to boot disks as well as instances. (Note: To set custom labels on PVs the csi-driver supports setting them via a StorageClass's `parameters`.

**Release note**:

```noteworthy user
Disks, boot disks, and instances will now share a common label 'k8s-cluster-name' that contains the shoot-ID.
Also, user-provided labels will now be added to boot disks as well.
```

### DIFF
--- a/pkg/controller/worker/machines.go
+++ b/pkg/controller/worker/machines.go
@@ -109,8 +109,9 @@ func (w *workerDelegate) generateMachineConfig(_ context.Context) error {
 			return err
 		}
 
-		arch := pointer.StringDeref(pool.Architecture, v1beta1constants.ArchitectureAMD64)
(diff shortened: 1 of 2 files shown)


kon-angelo (approved):
/lgtm

PR #671, opened 2023-12-12
Vendor gardener@v1.83.2

AndreasBurger:
/area open-source
/kind enhancement
/platform gcp

Revendors the extension to gardener `v1.83.2`

`-`

Need to double-check a few lines and (presumably) test -> Draft-PR
**Release note**:

```other operator
The following golang dependencies have been upgraded :
- `gardener/gardener`: `v1.81.6`->`v1.83.2`
```


AndreasBurger:
Local tests and integration-tests were successful so I'll promote this from draft-PR. Will investigate why _exactly_ the verify-step fails, as `make generate` does not create a diff for me locally.

AndreasBurger:
(Forcefully) updated with the expected chart, so everything should be in order.

pkg/controller/worker/machine_controller_manager.go
```diff
@@ -16,48 +16,14 @@ package worker
...
 	kutil "github.com/gardener/gardener/pkg/utils/kubernetes"
-	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
-	rbacv1 "k8s.io/api/rbac/v1"
 
-	"github.com/gardener/gardener-extension-provider-gcp/charts"
 	"github.com/gardener/gardener-extension-provider-gcp/pkg/gcp"
 )
 
-var (
```
kon-angelo, line 34: I think this file can be safely deleted. I did so in https://github.com/gardener/gardener-extension-provider-azure/pull/761 and not many things broke :)

kon-angelo (changes requested):

go.mod
```diff
@@ -8,9 +8,9 @@ require (
 	github.com/Masterminds/sprig v2.22.0+incompatible
 	github.com/ahmetb/gen-crd-api-reference-docs v0.3.0
 	github.com/coreos/go-systemd/v22 v22.5.0
-	github.com/gardener/etcd-druid v0.19.2
-	github.com/gardener/gardener v1.81.6
-	github.com/gardener/machine-controller-manager v0.48.1
+	github.com/gardener/etcd-druid v0.20.1
+	github.com/gardener/gardener v1.83.2
```
ialidzhikov, line 12: Is it possible to update to v1.83.3? In v1.83.3 an an issue in the extension library that prevents CPM for hibernated Shoots is resolved.

pkg/controller/worker/machine_controller_manager.go
AndreasBurger: Done.

go.mod
AndreasBurger: Done.

kon-angelo (approved):
/lgtm

PR #687, opened 2024-01-24
Make it possible to disable defaulting of storage classes

AndreasBurger:

/area storage
/kind enhancement
/platform gcp

With this change both the default StorageClass and VolumeSnapshot class can be configured to _not_ be tagged as default. This allows for non Gardener managed storage classes to be deployed and marked as default, as otherwise the ResourceManager would interfere here.

**Release note**:

```noteworthy operator
Tagging the 'default' StorageClass and VolumeSnapshotClass as default can now be disabled.
```


pkg/apis/gcp/v1alpha1/defaults.go
```diff
@@ -30,3 +30,16 @@ func SetDefaults_MachineImageVersion(obj *MachineImageVersion) {
...
 	}
 }
+
+// SetDefaults_Storage sets the defaults for the managed storage classes
+func SetDefaults_Storage(obj *Storage) {
+	defaultManagedDefaultStorageClass := true
+	defaultManagedDefaultVolumeSnapshotClass := true
+
+	if obj.ManagedDefaultStorageClass == nil {
+		obj.ManagedDefaultStorageClass = &defaultManagedDefaultStorageClass
```
kon-angelo, line 40: you can probably write it like `pointer.Bool(true)` which seems more succinct

pkg/controller/controlplane/valuesprovider.go
```diff
@@ -503,6 +503,34 @@ func getControlPlaneShootChartValues(
...
+
+	// Decode providerConfig
+	cpConfig := &apisgcp.ControlPlaneConfig{}
+	if cp.Spec.ProviderConfig != nil {
+		if _, _, err := vp.decoder.Decode(cp.Spec.ProviderConfig.Raw, nil, cpConfig); err != nil {
+			return nil, fmt.Errorf("could not decode providerConfig of controlplane '%s': %w", kutil.ObjectName(cp), err)
+		}
+	}
+
+	if cpConfig.Storage != nil && cpConfig.Storage.ManagedDefaultStorageClass != nil && cpConfig.Storage.ManagedDefaultVolumeSnapshotClass != nil {
```
kon-angelo, line 523: This is not an entirely correct assertion. You could for example not specify any value for the managed snapshot class in your configuration (and hence have it nil) but want to disable the default storageclass for example.

If you feel like the default storage class and snapshot class must be configured "together" and only like that, you should change the API spec to only expose one boolean instead

kon-angelo (changes requested):

kon-angelo (approved):
/lgtm

PR #691, opened 2024-01-29
deprecate mcm charts

kon-angelo:

/area control-plane
/kind cleanup
/platform gcp

Similar to https://github.com/gardener/gardener-extension-provider-alicloud/pull/687

Fixes #

**Release note**:

```other operator
The deprecated MCM deployment charts have now been removed.
```


rfranzke (commented):
There is more to be removed (see Alicloud PR) - what about those things?

kon-angelo:
@rfranzke I think the other things were removed during the vendoring PR. So the cmdline flags for the deployment should be already removed.

rfranzke:
OK great then :)

AndreasBurger (approved):

PR #723, opened 2024-04-04
Improve error msg when `serviceaccount.json` is in invalid json format

plkokanov:

/area usability ops-productivity
/kind enhancement
/platform gcp

This PR enhances the error which is returned if the gcp cloud provider secret contains a `serviceaccount.json` field which is in an invalid json format and cannot be unmarshalled. 
Without this, if someone tries to create a `SecretBinding` to such a secret containing an invalid `serviceaccount.json` the only error that is displayed is: 

```
Error from server: error when creating "secretbinding-cloudprovider.yaml": admission webhook "validator.admission-gcp.extensions.gardener.cloud" denied the request: unexpected end of JSON input
```

Now the error should be more descriptive pointing to its actual cause:
```
Error from server: error when creating "secretbinding-cloudprovider.yaml": admission webhook "validator.admission-gcp.extensions.gardener.cloud" denied the request: referenced secret foo/bar is not valid: could not get service account from "serviceaccount.json" field: failed to unmarshal json object: unexpected end of JSON input
```

Additionally, I noticed that the `readServiceAccountSecret` function is unused, so I removed it.

Fixes #

**Release note**:

```other user
An error text which better indicates the reason for the failure is displayed when a user tries to create a `SecretBinding` resource which references a `Secret` with a `serviceaccount.json` field in invalid json format.
```


AndreasBurger (approved):
lgtm

kon-angelo (approved):
/lgtm

PR #725, opened 2024-04-05
Proposal: Data disk restore from Image 

### LINKED ISSUE #695: Integrate vSMP MemeoryOne in GCP

/area usability
/area os
/kind enhancement

Gardener already supports multi-volume connection for machines used by MemeoryONE in AWS , to boot OS and MemeoryOne software on the same worker-node on separate dataVolumes. 
The same functionality needed in GCP for MemoryONE worker-nodes. 

I would post here a link to an epic for the similar functionality requested in AWS in the past. 
https://github.com/gardener/gardener/issues/2354

**Why is this needed**:

elankath:

/area documentation
/kind enhancement
/platform gcp

For support of vSMP One use-case, we need disk creation based off snapshot name.

Fixes #695

**Release note**:

```doc user
NONE
```


docs/proposals/datadisk-snapshot-restore.md
```diff
@@ -0,0 +1,83 @@
...
+```yaml
+providerConfig:
+  apiVersion: gcp.provider.extensions.gardener.cloud/v1alpha1
+  kind: WorkerConfig
+  volume:
+    interface: NVME
+    encryption: # optional, skipped detail here
+  dataVolumes: # <-- NEW SUB_SECTION
+    - name: vsmp1
+      snapshotName: snap-1234
```
kon-angelo, line 71: As with Azure, we need to see if it's better to support creating volumes from snapshots or images. In this case I would argue that the image is more important but we should check with the original stakeholder for the feature.

elankath: @kon-angelo  This is to make the feature in-line with our current AWS offering where we support restoring volumes from snapshots.

kon-angelo: I think in AWS you have no other option, but for "repeatably" creating a new disk seeded from a filesystem, an image is probably a better solution.

YakovKugel: As we decided during a call with @kon-angelo the target is to support images both for GCP and Azure as the data source for the volume. 
We understand that in AWS it is not possible. 
Whether or not to call the field "snap-name" or "image-name" is up to you.

elankath: @kon-angelo  should we have separate fields like `imageName`, `galleryImageName` ? Since the hierarchical `imageReference` would be formed differently in Azure in the MCM Azure provicer.

kon-angelo (approved):
/lgtm

However in practice I think we should go for `sourceImage` instead of `imageName` as that's how it is referred to by gcp client anyway

elankath:
> /lgtm
> 
> However in practice I think we should go for `sourceImage` instead of `imageName` as that's how it is referred to by gcp client anyway

Ok, you are right that the API field is `compute.AttachedDisk.SourceImage`  in the user doc it is called `image`

```
image
    Specifies the name of the image that the disk will be initialized with. A new disk will be created based on the given image. 
```

kon-angelo:
@elankath which documentation are you referring?  The go client has it defined as:
```
	SourceImage string `json:"sourceImage,omitempty"`
```
So the backend probably expects it as sourceImage.

kon-angelo (approved):
/lgtm

PR #730, opened 2024-04-12
Update golang versions in tests as well

AndreasBurger:
Updates the golang version used in the TM-tests to match the required version.


hebelsan (approved):
/lgtm

PR #751, opened 2024-04-30
Data volume source image

### LINKED ISSUE #695: Integrate vSMP MemeoryOne in GCP

hebelsan:

/area usability
/kind enhancement
/platform gcp

- Add support for specifying data volume sourceImage in the workerConfig

Fixes #695 

**Release note**:

```feature user
Add support for specifying data volume sourceImage in the workerConfig
```


kon-angelo:
@elankath I think you are correct. The way it is setup now *all* dataVolumes would use the same sourceImage, which is clearly not was intended.

hebelsan:
@elankath Could you please review again? :)

pkg/apis/gcp/validation/worker.go
```diff
@@ -120,22 +125,54 @@ func validateDataVolume(workerConfig *gcp.WorkerConfig, volume core.DataVolume,
...
 		return allErrs
 	}
-	if *volume.Type == worker.VolumeTypeScratch {
+
+	allErrs = append(allErrs, validateScratchDisk(*volume.Type, workerConfig)...)
+
+	return allErrs
+}
+
+func validateScratchDisk(volumeType string, workerConfig *gcp.WorkerConfig) field.ErrorList {
```
elankath, line 134: (apologies for delay in reviewing. was on leave.

Just to confirm :I assume this new validation routine was introduced since we didn't have one before to enforce this section in the usage docuement:

"If you attach the disk with SCRATCH type, either an NVMe interface or a SCSI interface must be specified. It is only meaningful to provide this volume interface if only SCRATCH data volumes are used."

pkg/controller/worker/machines.go
```diff
@@ -296,8 +305,8 @@ func createDiskSpec(size string, boot bool, machineImage, volumeType *string, wo
 		disk["labels"] = labels
 	}
 
-	if machineImage != nil {
-		disk["image"] = *machineImage
+	if image != nil {
```
elankath, line 308: Do we need any changes in `machine-controller-manager-provider-gcp` or just setting `image` is good enough ? Just wondering why we decided to choose `dataVolume.sourceImage` instead of simply `dataVolume.image` ?

pkg/apis/gcp/validation/worker.go
hebelsan: Previously, the validation for the SCRATCH disk type was present and has now been slightly refined. 
A new validation rule requires that the workerConfigDataVolume name must correspond to a dataVolume name.

pkg/controller/worker/machines.go
hebelsan: yes, nothing to change in the mcm.
we went for `sourceImage` because that's how it is referred to by gcp client.

elankath (approved):
thanks for the PR.

kon-angelo (approved):
/lgtm

PR #773, opened 2024-06-07
fix issue with updating an existing router with custom router adverti…

kon-angelo:
…sement

/area control-plane
/kind bug
/platform gcp

Fixes #

**Release note**:

```other operator
Fix an issue preventing the creation of NAT gateways on cloud routers with custom advertise mode
```

### DIFF
--- a/pkg/gcp/client/updater.go
+++ b/pkg/gcp/client/updater.go
@@ -209,8 +209,12 @@ func (u *updater) NAT(ctx context.Context, region string, router *compute.Router
 		return router, desired, nil
 	}
 
-	u.log.Info("updating router with NAT", "Name", router.Name)
-	router, err := u.compute.PatchRouter(ctx, region, router.Name, router)
+	routerPatch := compute.Router{
+		Name: router.Name,
+		Nats: router.Nats,
+	}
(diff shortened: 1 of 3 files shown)


AndreasBurger (approved):

pkg/gcp/client/updater.go
```diff
@@ -192,16 +192,20 @@ func (u *updater) NAT(ctx context.Context, region string, router *compute.Router
...
 	// not found case
 	if index < 0 {
-		router.Nats = append(router.Nats, desired)
+		routerPatch.Nats = append(router.Nats, desired)
 		modified = true
 	} else {
 		nats := router.Nats[index]
 		if !reflect.DeepEqual(nats, desired) {
 			modified = true
 			router.Nats[index] = desired
```
hebelsan, line 207: Is it ok to modify the original router object here?

kon-angelo: I will try to improve on that, but it doesn't make a difference. 

The router object is also replaced at the end of the patch call. In practice, the nats slice is modified in place due to a lack of deepCopy but since there is only one concurrent actor on the router at each time.

hebelsan (approved):
/lgtm

PR #781, opened 2024-06-11
Fix bug in scheduling-determination

AndreasBurger:

/area usability
/kind bug
/platform gcp

Fixes a bug that would cause invalid scheduling-configuration to be generated for GPU nodes. Also updates test-case so that both cases are covered.

**Release note**:

```bugfix user
Fixed a bug that could occur when configuring GPU-nodes, resulting in an invalid scheduling configuration.
```


kon-angelo (approved):
/lgtm

kon-angelo:
let's also cherry pick this once merged.

PR #818, opened 2024-08-07
Fix infra status egress CIDRs

hebelsan:

/area networking
/kind bug
/platform gcp

Fix a bug where we append instead of replace the egress CIDR infra status

Fixes #

**Release note**:

```bugfix operator
Fix a bug where we append instead of replace the egress CIDR infra status
```


kon-angelo (approved):
/lgtm

PR #826, opened 2024-08-23
[GEP-27] Use cloud profile bastion machine and image

hebelsan:

/area quality
/kind enhancement
/platform gcp

Instead of using a hardcoded bastion image we will use the image provided in the bastion section of the cloud profile.
See also [GEP proposal](https://github.com/gardener/gardener/pull/9935)

Fixes #

We need to wait until the bastion section is released with [gardener/gardener/types_cloudprofile.go](https://github.com/gardener/gardener/blob/master/pkg/apis/core/types_cloudprofile.go).

**Release note**:

```other operator
Create bastion vm from the info provided in the cloud profile bastion section
```


kon-angelo (commented):
Code is okay sans small nits but why isn't the `determineVMDetails` part of the g/g repo ? Just make it a library function that we can call from all providers. 

Is there any particular implementation details that change between the providers ?

pkg/controller/bastion/options.go
```diff
@@ -196,3 +193,45 @@ func FirewallEgressAllowOnlyResourceName(baseName string) string {
 func FirewallEgressDenyAllResourceName(baseName string) string {
 	return fmt.Sprintf("%s-deny-all", baseName)
 }
+
+func getCloudProfileConfig(cluster *extensions.Cluster) (*gcpv1alpha1.CloudProfileConfig, error) {
```
kon-angelo, line 197: Maybe this can be replaced with https://github.com/gardener/gardener-extension-provider-gcp/blob/07acf111fa7321939d63be6c36597cdacde0895c/pkg/apis/gcp/helper/scheme.go#L66 ?

kon-angelo (commented):
Just a nit about the CPConfig decoding

AndreasBurger (approved):

PR #831, opened 2024-09-02
Be more precise when updating firewall rules

AndreasBurger:

/area quality
/kind enhancement
/platform gcp

During reconciliation using the flow-based reconciler, we apply firewall rule updates by sending the updated rule wholesale. Terraform, on the other hand, does some calculations to reduce the update sent.
With this change we now also compute a firewall rule of minimal change and send this when applying, reducing the extension's footprint.

**Special notes for your reviewer**: I've opted for a somewhat simplified unit-test eschewing the complete ginkgo-stack.

**Release note**:

```noteworthy user
In order to reduce log events, only the minimal required changes will be made when applying firewall rules using the flow-reconciler. This matches the behaviour of the Terraform-reconciler.
```


pkg/controller/infrastructure/infraflow/ensure.go
```diff
@@ -320,12 +320,20 @@ func (fctx *FlowContext) ensureFirewallRules(ctx context.Context) error {
...
-			continue
-		}
-		_, err = fctx.updater.Firewall(ctx, rule)
-		if err != nil {
-			log.Info(fmt.Sprintf("failed to update firewall %s rule: %v", rule.Name, err))
-			return err
+		} else {
+			updateRule, err := firewallUpdate(gcpRule, rule)
+			if err != nil {
+				log.Info(fmt.Sprintf("failed calculate update for firewall %s rule: %v", rule.Name, err))
```
AndreasBurger, line 326:
```suggestion
				log.Info(fmt.Sprintf("failed to calculate update for firewall %s rule: %v", rule.Name, err))
```
I seem to have accidentally a word.

Will be fixed.

pkg/controller/infrastructure/infraflow/ensure_utils.go
```diff
@@ -296,3 +299,145 @@ func isUserRouter(config *gcp.InfrastructureConfig) bool {
 func isUserVPC(config *gcp.InfrastructureConfig) bool {
 	return config.Networks.VPC != nil && len(config.Networks.VPC.Name) > 0
 }
+
+// isSimilar return whether to slices are similar using reflect.DeepEqual
```
hebelsan, line 303: small typo: your probably want to say `two slices`

pkg/controller/infrastructure/infraflow/ensure_utils.go
```diff
@@ -296,3 +299,145 @@ func isUserRouter(config *gcp.InfrastructureConfig) bool {
...
+		}
+	}
+	return true
+}
+
+// firewallUpdate returns a firewall rule that reflects the differences from 'left' towards 'right'.
+//
+// The intended use for the returned firewall rule is to serve as a minimal patch when updating
+// firewall rules on gcp.
+func firewallUpdate(left, right *compute.Firewall) (*compute.Firewall, error) {
```
hebelsan, line 323: Wouldn't `current` and `new` be more appropriate instead of `left` and `right`?

pkg/controller/infrastructure/infraflow/ensure.go
```diff
@@ -320,12 +320,20 @@ func (fctx *FlowContext) ensureFirewallRules(ctx context.Context) error {
...
+				return err
+			}
+			if updateRule != nil {
+				if _, err = fctx.updater.Firewall(ctx, updateRule); err != nil {
```
hebelsan, line 330: Is it really necessary to compute the diff and update the rule with this diff? Couldn't we just check if we need to update, and if so use `rule` instead of `updateRule`?

hebelsan (commented):
Do we need to adjust other providers as well?

pkg/controller/infrastructure/infraflow/ensure_utils.go
```diff
@@ -296,3 +299,145 @@ func isUserRouter(config *gcp.InfrastructureConfig) bool {
...
+	if left == nil && right != nil {
+		// everything is right, return right
+		return right, nil
+	}
+	if right == nil && left != nil {
+		// strictly speaking, the update here would be an empty Firewall-object with its
+		// ForceSendFields set to all fields. In the intended context of this function,
+		// just return nil signalling that there is nothing to do
+		return nil, nil
+	}
```
kon-angelo, line 337:
```
	if left == nil {
		return right, nil
	}
	if right == nil || right == left {
		return nil, nil
	}
```

Although I would consider any case where the `right == nil` either a delete event or an error. I don't think you can even reach this func with a `left == nil` so maybe that's an error too ?

pkg/controller/infrastructure/infraflow/ensure_utils.go
AndreasBurger: It previously was `old` and `new`, but the linter complains about using `new` (rightfully so, as it shadows a golang keyword)

hebelsan: right, makes sense.
maybe then oldF and newF?
I'm not a big fan of left and right :D

hebelsan: well, only if the comment below is not true.

pkg/controller/infrastructure/infraflow/ensure.go
AndreasBurger: I'm afraid so.
The entire point of this PR is to _not_ use `rule` as it has _some_ fields set to non-zero values. These fields will cause noop-updates on the GCP side if used in the update call on our side. GCP does not check whether an update is an actual change server-side. The only way around that is to use a firewall-rule with zeroed fields, which is `updateRule`.

pkg/controller/infrastructure/infraflow/ensure_utils.go
```diff
@@ -296,3 +299,145 @@ func isUserRouter(config *gcp.InfrastructureConfig) bool {
...
+	if right.Network != "" && left.Network != right.Network {
+		diff["network"] = right.Network
+	}
+	if right.Priority != 0 && left.Priority != right.Priority {
+		diff["priority"] = right.Priority
+	}
+	if right.SelfLink != "" && left.SelfLink != right.SelfLink {
+		diff["selfLink"] = right.SelfLink
+	}
```
kon-angelo, line 400: why isn't checking just the inequality enough ?

AndreasBurger:
> Do we need to adjust other providers as well?

No, this is both a unique problem to GCP and I think the other SDKs either don't allow this optimization or it is done for you on the backend

pkg/controller/infrastructure/infraflow/ensure_utils.go
```diff
@@ -296,3 +299,145 @@ func isUserRouter(config *gcp.InfrastructureConfig) bool {
...
+		}
+	}
+	if len(right.Denied) != 0 {
+		if len(left.Denied) == 0 || !isSimilar(left.Denied, right.Denied) {
+			diff["denied"] = right.Denied
+		}
+	}
+	if right.Description != "" && left.Description != right.Description {
+		diff["description"] = right.Description
+	}
```
kon-angelo, line 360: careful here, the description is immutable. We would have to delete and then recreate the rule.

pkg/controller/infrastructure/infraflow/ensure.go
kon-angelo: Not sure I understand the response here. If you detected that the `rule` has not changed, you would skip the update call altogether instead of doing it. If you will do an update anyway, why bother being precise about the fields ?

pkg/controller/infrastructure/infraflow/ensure_utils.go
```diff
@@ -296,3 +299,145 @@ func isUserRouter(config *gcp.InfrastructureConfig) bool {
...
+	if len(s1) != len(s2) {
+		return false
+	}
+	for _, v := range s1 {
+		if !slices.ContainsFunc(s2, func(e T) bool { return reflect.DeepEqual(v, e) }) {
+			return false
+		}
+	}
+	return true
+}
```
kon-angelo, line 317: Maybe `isEquivalent` is better here ?

pkg/controller/infrastructure/infraflow/ensure_utils.go
AndreasBurger: These were not reduced for legibility.

Basically these lines attempt to handle the obvious default cases. Thinking about it, I don't think this is the way to go for a function with such a limited scope. I'll replace it with checks for `nil` and require two non-nil inputs.

pkg/controller/infrastructure/infraflow/ensure.go
```diff
@@ -320,12 +320,20 @@ func (fctx *FlowContext) ensureFirewallRules(ctx context.Context) error {
 				log.Info(fmt.Sprintf("failed to create firewall %s rule: %v", rule.Name, err))
 				return err
 			}
-			continue
-		}
-		_, err = fctx.updater.Firewall(ctx, rule)
-		if err != nil {
-			log.Info(fmt.Sprintf("failed to update firewall %s rule: %v", rule.Name, err))
-			return err
+		} else {
```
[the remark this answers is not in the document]
AndreasBurger: I prefer if/else to if/continue for loops, so this will probably remain.

Not sure about the nested one, though. It's only there for the logging

pkg/controller/infrastructure/infraflow/ensure_utils.go
```diff
@@ -296,3 +298,98 @@ func isUserRouter(config *gcp.InfrastructureConfig) bool {
...
+
+// shouldUpdate returns a boolean indicating whether there is a change between the two given rules
+// that would necessitate an update
+func shouldUpdate(oldRule, newRule *compute.Firewall) bool {
```
kon-angelo, line 320: similar logic we (aka "I") hide in the `updater` see https://github.com/gardener/gardener-extension-provider-gcp/blob/450a0e7a52f0f437ca33e426d909b06f03c0b72e/pkg/gcp/client/updater.go#L158. I am kinda convinced that the updater can contain this logic, although I would probably discuss if the client package is the best home for the updater

pkg/controller/infrastructure/infraflow/ensure_utils.go
```diff
@@ -296,3 +298,98 @@ func isUserRouter(config *gcp.InfrastructureConfig) bool {
...
+	}
+	if oldRule == nil || newRule == nil {
+		return true
+	}
+
+	// Check everything except the immutable 'Name', 'Description', and 'SelfLink'.
+	// Also CreationTimestamp is ignored.
+
+	if len(newRule.Allowed) != 0 {
+		if len(oldRule.Allowed) == 0 || !isEquivalent(oldRule.Allowed, newRule.Allowed) {
```
kon-angelo, line 332: `isEquivalent` already checks the `len` right ?

hebelsan (approved):
/lgtm

kon-angelo (approved):
/lgtm

PR #835, opened 2024-09-04
Unify gcp clients

hebelsan:

/area quality
/kind enhancement
/platform gcp

In provider-gcp we have 2 client implementations which is redundant.
This PR combines them into one.

Fixes #

**Release note**:

```developer operator
Combine two gcp client implementations (of which one is redundant) into one
```


pkg/controller/bastion/actuator.go
```diff
@@ -42,31 +43,15 @@ func newActuator(mgr manager.Manager) bastion.Actuator {
...
-	return firewall, nil
+func getBastionInstance(ctx context.Context, client gcpclient.ComputeClient, opt *Options) (*compute.Instance, error) {
+	instance, err := client.GetInstance(ctx, opt.Zone, opt.BastionInstanceName)
+	return instance, gcpclient.IgnoreNotFoundError(err)
 }
 
-func createFirewallRuleIfNotExist(ctx context.Context, log logr.Logger, gcpclient gcpclient.Interface, opt *Options, firewallRule *compute.Firewall) error {
-	if _, err := gcpclient.Firewalls().Insert(opt.ProjectID, firewallRule).Context(ctx).Do(); err != nil {
-		if googleError, ok := err.(*googleapi.Error); ok && googleError.Code == http.StatusConflict {
+func createFirewallRuleIfNotExist(ctx context.Context, log logr.Logger, client gcpclient.ComputeClient, firewallRule *compute.Firewall) error {
```
kon-angelo, line 51: (okay it was there) but do we need to the if-not-exists part ? We are not doing a get here right ?

pkg/controller/bastion/actuator.go
```diff
@@ -42,31 +43,15 @@ func newActuator(mgr manager.Manager) bastion.Actuator {
...
+	if _, err := client.InsertFirewallRule(ctx, firewallRule); err != nil {
+		var googleError *googleapi.Error
+		if errors.As(err, &googleError) && googleError.Code == http.StatusConflict {
```
kon-angelo, line 54: maybe you can use https://github.com/gardener/gardener-extension-provider-gcp/blob/c73f2b87c2b67c9f5a5907113adb47b478342881/pkg/gcp/client/errors.go#L15

pkg/controller/bastion/actuator.go
hebelsan: Good point, I'n not 100% sure.
Wouldn't we catch it because we check for http.StatusConflict upon inserting?

kon-angelo (approved):
/lgtm

PR #842, opened 2024-09-10
Update CCM for k8s 1.31

### LINKED ISSUE #679: Consider switching to the out-of-tree GCP cloud-controller-manager

/area control-plane
/kind enhancement
/platform gcp

Although https://github.com/kubernetes/cloud-provider-gcp does not have any github releases, its container images are already available. From https://github.com/kubernetes/cloud-provider-gcp/issues/289#issuecomment-1889176432:
> The CCM image for Kubernetes 1.29.0 is not yet available / published, there's only
> 
> * [v26.2.4](https://explore.ggcr.dev/?image=registry.k8s.io%2Fcloud-provider-gcp%2Fcloud-controller-manager:v26.2.4)
> * [v27.1.6](https://explore.ggcr.dev/?image=registry.k8s.io%2Fcloud-provider-gcp%2Fcloud-controller-manager:v27.1.6)
> * [v28.2.1](https://explore.ggcr.dev/?image=registry.k8s.io%2Fcloud-provider-gcp%2Fcloud-controller-manager:v28.2.1)

Tagging it is a little bit odd: https://github.com/kubernetes/cloud-provider-gcp/issues/289#issuecomment-1540327387
`@v26.x.y` corresponds to K8s v1.26, `@v27.x.y` corresponds to K8s 1.27, etc.

And again:
> We can check the compatibility with the in-tree CCM and, if possible, switch to the out-of-tree GCP CCM. We could do this by deploying the OOT CCM staring given K8s version (for example starting K8s 1.29).

**Why is this needed**:
To get rid of our last active in-tree CCM fork (https://github.com/gardener/cloud-provider-gcp) and all the maintenance effort related to it.

AndreasBurger:

/area control-plane
/kind cleanup
/platform gcp

Updates the CCM to be used for kubernetes `1.31.0` or greater to the one provided by GCP.

Fixes #679

see inline comment

**Release note**:

```noteworthy operator
Switch to upstream CCM for kubernetes versions greater than `1.31.0`
```

### DIFF
--- a/charts/internal/seed-controlplane/charts/cloud-controller-manager/templates/cloud-controller-manager.yaml
+++ b/charts/internal/seed-controlplane/charts/cloud-controller-manager/templates/cloud-controller-manager.yaml
@@ -37,6 +37,9 @@ spec:
       - name: gcp-cloud-controller-manager
         image: {{ index .Values.images "cloud-controller-manager" }}
         imagePullPolicy: IfNotPresent
+        {{- if semverCompare ">= 1.31" .Values.kubernetesVersion }}
+        command: ["/cloud-controller-manager"]
+        {{- end}}
         args:
         - --allocate-node-cidrs=true
         - --cloud-provider=gce
--- a/charts/internal/seed-controlplane/charts/cloud-controller-manager/values.yaml
+++ b/charts/internal/seed-controlplane/charts/cloud-controller-manager/values.yaml
@@ -9,7 +9,7 @@ images:
   cloud-controller-manager: image-repository:image-tag
 resources:
   requests:
-    cpu: 100m 
+    cpu: 100m
     memory: 75Mi
 tlsCipherSuites: []
 secrets:
--- a/imagevector/images.yaml
+++ b/imagevector/images.yaml
@@ -86,7 +86,21 @@ images:
   sourceRepository: github.com/gardener/cloud-provider-gcp
   repository: europe-docker.pkg.dev/gardener-project/releases/kubernetes/cloud-provider-gcp
   tag: "v1.30.0"
-  targetVersion: ">= 1.30"
+  targetVersion: "1.30.x"
+  labels:
+  - name: 'gardener.cloud/cve-categorisation'
+    value:
+      network_exposure: 'protected'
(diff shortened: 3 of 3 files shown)


charts/internal/seed-controlplane/charts/cloud-controller-manager/templates/cloud-controller-manager.yaml
```diff
@@ -37,6 +37,9 @@ spec:
       - name: gcp-cloud-controller-manager
         image: {{ index .Values.images "cloud-controller-manager" }}
         imagePullPolicy: IfNotPresent
+        {{- if le (semver .Values.kubernetesVersion | (semver "1.31.0").Compare) 0 }}
+        command: ["/cloud-controller-manager"]
+        {{- end}}
```
AndreasBurger, line 42: Not entirely sure whether that is the way to go. This would work, but it is a bit unwieldy and we should remove the condition once older versions drop out of support.

Since our image uses `CMD` and the upstream one uses `ENTRYPOINT` to specify the binary to run during the build process, we could just use `command` for both. I suppose that would restart all CCMs if we update. Not sure if that is actually a problem.

AndreasBurger:
verify failing due to golang `1.23` related stuff. Will rebase once fix is on master.

AndreasBurger:
Rebased.

charts/internal/seed-controlplane/charts/cloud-controller-manager/templates/cloud-controller-manager.yaml
```diff
@@ -37,6 +37,9 @@ spec:
       - name: gcp-cloud-controller-manager
         image: {{ index .Values.images "cloud-controller-manager" }}
         imagePullPolicy: IfNotPresent
+        {{- if le (semver .Values.kubernetesVersion | (semver "1.31.0").Compare) 0 }}
```
kon-angelo, line 40: why not:
```
{{- if semverCompare ">= 1.31" .Values.kubernetesVersion }}
```

kon-angelo (approved):
/lgtm

PR #862, opened 2024-10-09
Add support for VolumeAttributeClasses

AndreasBurger:

/area storage
/kind enhancement
/platform gcp

With this PR support for [volume attributes classes](https://kubernetes.io/docs/concepts/storage/volume-attributes-classes/) is added by allowing to configure the necessary parameters for it to work.

Fixes #

**Release note**:

```other operator

```


AndreasBurger:
Currently WIP; with the (currently hardcoded) changes it should work, but I do not have a way to get a 1.31 cluster to fully test yet. So, open todos:
- test
- make usable/configurable via shoot-manifest

kon-angelo (changes requested):

charts/internal/seed-controlplane/charts/csi-driver-controller/templates/csi-driver-controller.yaml
```diff
@@ -49,6 +49,12 @@ spec:
         - --logtostderr
         - --v=3
         - --enable-storage-pools
+        {{- if (((.Values.csiDriver).storage).supportsDynamicIopsProvisioning) }}
```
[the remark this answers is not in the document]
AndreasBurger: Looking at the documentation, I'm still not sure that `index .Values csiDriver storage supportsDynamicIopsProvisioning` would behave the same with regards to absent values. I'll try to get around to testing it tomorrow.

If there is no difference there is no reason 😅.

kon-angelo (approved):
/lgtm

pkg/controller/controlplane/valuesprovider.go
```diff
@@ -477,7 +477,13 @@ func getCSIControllerChartValues(
...
 	}
 
-	if semver.Compare(cluster.Shoot.Spec.Kubernetes.Version, "1.30.0") > 0 {
+	// golang mod's semver considers only semvers starting with 'v' to be proper semver, so we need to do some checking/converting here.
+	// On the bright side, it can handle the shoot-spec-supported case of partial semvers, e.g.: (v)1.31
+	k8sVersion := cluster.Shoot.Spec.Kubernetes.Version
+	if !strings.HasPrefix(k8sVersion, "v") {
+		k8sVersion = "v" + k8sVersion
+	}
+	if semver.Compare(k8sVersion, "v1.30.0") > 0 {
```
hebelsan, line 486: Instead of the hack we maybe could make use `versionutils.ConstraintK8sLess130.Check(k8sVersion)` like in this example: https://github.com/gardener/gardener-extension-provider-gcp/blob/07acf111fa7321939d63be6c36597cdacde0895c/pkg/webhook/controlplane/ensurer.go#L183

kon-angelo: 👍 on the suggestion

AndreasBurger:
/hold

now requires gardener v1.106

kon-angelo (approved):
/lgtm

PR #876, opened 2024-10-23
remove outdated external access rule

kon-angelo:

/area control-plane
/kind enhancement
/platform gcp

Fixes #

**Release note**:

```other operator
Remove redundant firewall rule `*-allow-external-access`
```


ScheererJ (approved):
/lgtm

As discussed offline, I think this should work well without this rule.

PR #911, opened 2024-11-25
csi-driver-node: set GOMAXPROCS

Garfield96:

/area performance
/kind enhancement
/platform gcp

Golang creates internal data structures based on the number of available CPUs. As a result, the memory consumption increases on larger machines. However, these data structures would be only needed if the load would scale with the size of the nodes. This isn't the case for CSI-drivers, since the number of volumes doesn't scale with the node size. Therefore, this change overwrites the automatically detected number of CPUs and sets it to 2 using the [GOMAXPROCS](https://pkg.go.dev/runtime#GOMAXPROCS) environment variable. Therefore, memory consumption is stable across all node sizes. As a result, the VPA doesn't have to scale up the resources if a cluster uses many large nodes, which in the worst case could result in pod eviction.

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/charts/internal/shoot-system-components/charts/csi-driver-node/templates/daemonset.yaml
+++ b/charts/internal/shoot-system-components/charts/csi-driver-node/templates/daemonset.yaml
@@ -41,6 +41,7 @@ spec:
         args:
         - --endpoint=$(CSI_ENDPOINT)
         - --run-controller-service=false
+        - --maxprocs=2 # https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/issues/968
         - --logtostderr
         - --v=5
         env:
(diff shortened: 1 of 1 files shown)


gardener-robot-ci-2:
Thank you @Garfield96 for your contribution. Before I can start building your PR, a member of the organization must set the required label(s) {'reviewed/ok-to-test'}. Once started, you can check the build status in the PR checks section below.

kon-angelo:
@Garfield96 Thank you for the find. Quite interesting indeed. 

Small nit: I think the original upstream issue is https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/issues/968, which may be interesting. Could please add a comment with this issue and I believe it's okay to proceed.

kon-angelo:
Also based on the [upstream PR](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/969/files), maybe the idiomatic way is to use the provided csi flag

AndreasBurger (approved):
nice find, lgtm

PR #912, opened 2024-11-26
csi-driver-node: enable readOnlyRootFilesystem

Garfield96:

/area robustness
/area security
/kind enhancement
/platform gcp

This PR makes the container root filesystem of all containers of the csi-driver-node pod read-only. This improves the robustness of the cluster, since a faulty application cannot write data in an uncontrolled fashion, which in the worst case can affect all pods on the host node. It also hardens the application, since many exploits are based on writing a script or executable to disk and executing it afterwards. This is no longer possible after this change.

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/charts/internal/shoot-system-components/charts/csi-driver-node/templates/daemonset.yaml
+++ b/charts/internal/shoot-system-components/charts/csi-driver-node/templates/daemonset.yaml
@@ -52,6 +52,7 @@ spec:
 {{- end }}
         securityContext:
           privileged: true
+          readOnlyRootFilesystem: true
         ports:
         - name: healthz
           containerPort: 9808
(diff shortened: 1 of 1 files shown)


gardener-robot-ci-3:
Thank you @Garfield96 for your contribution. Before I can start building your PR, a member of the organization must set the required label(s) {'reviewed/ok-to-test'}. Once started, you can check the build status in the PR checks section below.

kon-angelo:
> since a faulty application cannot write data in an uncontrolled fashion, which in the worst case can affect all pods on the host node.

@Garfield96 can you give some more information about the concrete issue we are solving here? I don't think that setting this hurts us, but the upstream deploy charts neither use it, nor recommend it.

Garfield96:
Hi @kon-angelo,

> since a faulty application cannot write data in an uncontrolled fashion, which in the worst case can affect all pods on the host node.

we did not observe this behavior for the csi-driver or its sidecars, but if there is no ephemeral storage limit, a container can write to the root file system, until the host disk is filled up, which can make the whole node instable. This change is primarily meant as a security improvement. I'll create the same change in the upstream project.

AndreasBurger (approved):

PR #929, opened 2024-12-18
Remove VPA for the csi-driver-node DaemonSet

ialidzhikov:

/area auto-scaling
/kind enhancement
/platform gcp

On our side, we are facing many occurrences of Pods with volumes scheduled on Nodes which volume attachment limit is exhausted. Long story short, on csi-driver-node Pod termination, it removes the volume attachment limit from the CSINode  object and re-adds it on start. In the time frame when no csi-driver-node Pod is running for the Node, the CSINode object contains no limit and kube-scheduler assumes that the Node does not have a volume attachment limit and allows Pods with volumes to be scheduled on that Node. At the end, workload Pod(s) hang in `ContainerCreating` state because their volume cannot be attached to the Node. For more details, see https://github.com/kubernetes/kubernetes/issues/126921.

We noticed that csi-driver-node Pod restarts are caused by VPA evictions. To mitigate the above issue, we decided to disable VPA for the csi-driver-node as it does not need VPA and does not benefit from it.

Internal usage data for the csi-driver-node containers:
- `csi-driver`
  * memory usage min: 11.4MiB
  * memory usage avg: 21.0MiB
  * memory usage max: 58.6MiB
  * cpu usage min: 0.142m
  * cpu usage avg: 0.767m
  * cpu usage max: 32.2m
- `csi-liveness-probe`
  * memory usage min: 8.84MiB
  * memory usage avg: 14.2MiB
  * memory usage max: 22.8MiB
  * cpu usage min: 0.176m
  * cpu usage avg: 0.264m
  * cpu usage max: 3.2m
- `csi-node-driver-registrar`
  * memory usage min: 7.15MiB
  * memory usage avg: 11.8MiB
  * memory usage max: 23.4MiB
  * cpu usage min: 0.0148m
  * cpu usage avg: 0.0649m
  * cpu usage max: 2.57m

N/A

I haven't updated the resource requests https://github.com/gardener/gardener-extension-provider-gcp/blob/bc9cfc1eaefaeb02bfb4eeec9c78cca173e8b063/charts/internal/shoot-system-components/charts/csi-driver-node/values.yaml#L15-L27 as they are close to the internal recommendations.

The internal recommendations are:
- `csi-driver-node/csi-driver` container:  5m and 41Mi
- `csi-driver-node/csi-liveness-probe` container:  5m and 22.9Mi
- `csi-driver-node/csi-node-driver-registrar`: 5m and 21Mi

**Release note**:

```noteworthy user
The kube-system/csi-driver-node DaemonSet is no longer scaled by a VerticalPodAutoscaler as it does not really benefit from it. Removing VerticalPodAutoscaler for that component will reduce unnecessary evictions by VPA and will be a mitigation for https://issues.k8s.io/126921.
```


hebelsan (commented):
/lgtm

AndreasBurger (approved):

PR #937, opened 2025-01-07
Disable --allocate-node-cidrs in `kube-controller-manager` for kubernetes >= v1.31.

hebelsan:

/area control-plane
/area networking
/area robustness
/kind bug
/platform gcp

The out-of-tree `cloud-controller-manager` used with kubernetes v1.31 and above has a real IPAM instead of the no-op one from the previously used legacy in-tree `cloud-controller-manager`. This means that `--allocate-node-cidrs=true` has an effect on the new implementation. Unfortunately, we cannot easily disable it as well because then `cloud-controller-manager` does not start with the error `the AllocateNodeCIDRs is not enabled`.
Therefore, this change disables the `--allocate-node-cidrs` option in `kube-controller-manager` for kubernetes clusters >= v1.31 so that only a single controller acts on new nodes and thereby pod IPs will not be allocated in a potentially duplicate manner.

Fixes #

Having two controllers (kube-controller-manager and cloud-controller-manager) try to allocate pod ranges for nodes in parallel can easily lead to duplicate pod ranges and thereby duplicate pod IPs. The resulting behaviour is more or less undefined. This happens more frequently in case the node churn rate is high, i.e. if a lot of nodes are joining a cluster more or less around the same point in time.

**Release note**:

```other operator
Disable --allocate-node-cidrs in kube-controller-manager for kubernetes >= 1.31 as cloud-controller-manager takes over this responsibility.
```

### DIFF
--- a/pkg/webhook/controlplane/ensurer.go
+++ b/pkg/webhook/controlplane/ensurer.go
@@ -8,6 +8,7 @@ import (
 	"bytes"
 	"context"
 	"regexp"
+	"strconv"
 
 	"github.com/Masterminds/semver/v3"
 	"github.com/coreos/go-systemd/v22/unit"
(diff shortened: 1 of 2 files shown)


LucaBernstein (approved):
Thank you for the backport.
/lgtm

kon-angelo:
I will write it here for future reference in case this is needed, but you should be able to disable the ipam on the CCM with:

```
- --controllers=*,-nodeipam
- --allocate-node-cidrs=false
```

PR #985, opened 2025-03-03
fix an issue where wrong firewall rules wrongfully deleted

kon-angelo:

/area control-plane
/kind bug
/platform gcp

Fixes #

**Release note**:

```other operator
Fix a bug causing the deletion of firewall rules from different shoots if the shoot names had identical prefixes
```


AndreasBurger (approved):
lgtm

hebelsan (commented):
Just to be sure, we don't want to delete rules that fall under "FirewallRuleAllowExternalName" right?

kon-angelo:
Nice observation @hebelsan. Well, we delete that during normal reconciliation, but it could be that a cluster that has not reconciled long enough and goes straight into deletion still has that rule. I will update the PR

AndreasBurger (approved):

PR #986, opened 2025-03-04
remove dual-stack field from hash

DockToFuture:

/area networking
/kind enhancement
/platform gcp

Remove dual-stack field from hash to avoid directly triggering a node rollout during migration.

Fixes #

**Release note**:

```other operator
none
```


kon-angelo (approved):
/lgtm

PR #988, opened 2025-03-05
fix flow log and log config dependency

DockToFuture:

/area networking
/kind bug
/platform gcp

Fixes a dependency bug of flow log and log config.
Fixes #

**Release note**:

```other operator
Fix a dependency bug of flow log and log config.
```


hebelsan (approved):
/lgtm

PR #999, opened 2025-03-10
Restrict dual-stack clusters on GCP to Kubernetes v1.31 and above.

ScheererJ:

/area networking
/kind enhancement
/platform gcp

Restrict dual-stack clusters on GCP to Kubernetes v1.31 and above.

The new cloud controller manager release is a precondition so that the corresponding IPAM component can be used, which assigns a part of the pod range to a node as secondary IP range. This is only available for Kubernetes v1.31 as the old cloud controller manager is used for previous releases.
This change prevents the creation of clusters that will fail anyway.

Fixes #

**Release note**:

```other operator
Dual-stack clusters are only possible with Kubernetes v1.31 and above.
```

### DIFF
--- a/pkg/admission/validator/shoot.go
+++ b/pkg/admission/validator/shoot.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"reflect"
 
+	"github.com/Masterminds/semver/v3"
 	extensionswebhook "github.com/gardener/gardener/extensions/pkg/webhook"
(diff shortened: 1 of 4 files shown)


kon-angelo (dismissed):
/lgtm

DockToFuture (approved):
/lgtm

kon-angelo (approved):

PR #1000, opened 2025-03-10
Only add IPv6 egress CIDRs if static NAT gateways are configured for dual-stack clusters.

ScheererJ:

/area networking
/kind enhancement
/platform gcp

Only add IPv6 egress CIDRs if static NAT gateways are configured for dual-stack clusters.

The egress CIDRs for IPv6 are static and could be provided always. However, this could be confusing if the IPv4 NAT gateway was not configured to use static IPs. Then, the egress CIDRs would only contain IPv6 addresses, but the cluster may access external resources also via IPv4. Hence, using this as the base for IP allowlisting is not possible.

Therefore, this change does not add the egress CIDRs for IPv6 in case the NAT gateway is configured with dynamic IPs.

Fixes #

**Release note**:

```other operator
The egress CIDRs will be provided for dual-stack clusters only if both IPv4 and IPv6 egress CIDRs are known, i.e. the NAT gateway needs to be configured with static IPs.
```


kon-angelo (approved):
/lgtm

PR #1010, opened 2025-03-18
add ipv4 to dual stack migration

DockToFuture:

/area networking
/kind enhancement
/platform gcp

To migrate shoots from single stack to dual-stack networking, we track the current migration state in the shoot state. This PR adds checks to read the respective constraint in the shoot and configure components accordingly.
It removes the pod routes from the VPC routing table and adds alias ip ranges to the NIC for existing nodes, it stores the the deleted routes in the whiteboard until the corresponding alias ip ranges were added.

Fixes #

**Release note**:

```other operator
Add support for single-stack to dual-stack network migration.
```


pkg/gcp/client/compute.go
```diff
@@ -409,6 +414,45 @@ func (c *computeClient) ListRoutes(ctx context.Context, opts RouteListOpts) ([]*
...
+	var op *compute.Operation
+	for i := 0; i < maxRetries; i++ {
+		op, err = c.service.Instances.UpdateNetworkInterface(c.projectID, route.Zone, route.InstanceName, instance.NetworkInterfaces[0].Name, instance.NetworkInterfaces[0]).Context(ctx).Do()
+		if err == nil {
+			log.Info("Operation to update network interface started: ", "opName", op.Name, "name", instance.NetworkInterfaces[0].Name)
+			return nil
+		}
+
+		log.Info("Failed to update network interface", "attempt", strconv.Itoa(i), "name", instance.NetworkInterfaces[0].Name, "error", err)
+		time.Sleep(initialBackoff * time.Duration(backoffFactor^i))
```
hebelsan, line 449: This sleep function looks a bit weird.
If I'm not mistaken is `backoffFactor^i`  a bitwise XOR, not exponentiation.

pkg/gcp/client/compute.go
```diff
@@ -409,6 +414,45 @@ func (c *computeClient) ListRoutes(ctx context.Context, opts RouteListOpts) ([]*
 	return res, nil
 }
 
+func (c *computeClient) InsertAliasIPRoute(ctx context.Context, route apisgcp.Route, defaultSecondarySubnetName string) error {
+
+	maxRetries := 5
+	initialBackoff := 1 * time.Second
+	backoffFactor := 2
```
hebelsan, line 421: These are all constants

pkg/apis/gcp/types_infrastructure.go
```diff
@@ -170,4 +170,17 @@ type InfrastructureState struct {
 	// Data is map to store things.
 	// +optional
 	Data map[string]string
+	// Routes contains information about cluster routes
+	// +optional
+	Routes []Route
+}
+
+// Route is a structure containing information about the routes.
+type Route struct {
```
hebelsan, line 179: It reads a bit odd to suddenly store something of the infra state in a new dedicated struct, instead of putting it in the already existing Data field.
Is there a certain reason for not using the Data field and creating new key here?https://github.com/gardener/gardener-extension-provider-gcp/blob/64c11a37c8a6c515d447aee9a18b23ee4c388680/pkg/controller/infrastructure/infraflow/reconciler.go#L30

hebelsan: That's true.
Maybe `NetworkStatus` would be better?
https://github.com/gardener/gardener-extension-provider-gcp/blob/6b64e745e6709b5014c977e76f262f35b9aee106/pkg/apis/gcp/v1alpha1/types_infrastructure.go#L58

DockToFuture: Not sure about this, I thought the InfrastructureState struct was created to contain such data.

hebelsan (dismissed):
/lgtm

pkg/apis/gcp/v1alpha1/types_infrastructure.go
```diff
@@ -179,4 +181,7 @@ type InfrastructureState struct {
 	// Data is map to store things.
 	// +optional
 	Data map[string]string `json:"data,omitempty"`
+	// Routes contains information about cluster routes
+	// +optional
+	Routes []apisgcp.Route `json:"routes,omitempty"`
```
kon-angelo, line 186: For consistency I would mirror the struct here instead of importing the one from the internal package.

pkg/gcp/client/compute.go
```diff
@@ -409,6 +415,45 @@ func (c *computeClient) ListRoutes(ctx context.Context, opts RouteListOpts) ([]*
...
+	instance.NetworkInterfaces[0].AliasIpRanges = append(instance.NetworkInterfaces[0].AliasIpRanges, aliasIpRange)
+
+	// Retry logic for updating the instance
+	var op *compute.Operation
+	for i := 0; i < maxRetries; i++ {
+		op, err = c.service.Instances.UpdateNetworkInterface(c.projectID, route.Zone, route.InstanceName, instance.NetworkInterfaces[0].Name, instance.NetworkInterfaces[0]).Context(ctx).Do()
+		if err == nil {
+			log.Info("Operation to update network interface started: ", "opName", op.Name, "name", instance.NetworkInterfaces[0].Name)
+			return nil
+		}
```
kon-angelo, line 447: - Do we need a retry for this operation ?
- Do you not need to check if the operation completes successfully ? Here you only check that the operation is queued.
- I would say that maybe extract the exp.Retry in it's own func instead.

pkg/controller/infrastructure/infraflow/ensure.go
```diff
@@ -163,12 +164,31 @@ func (fctx *FlowContext) ensureKubernetesRoutesCleanup(ctx context.Context) erro
 		return err
 	}
 
-	if err = infrastructure.DeleteRoutes(ctx, fctx.computeClient, cloudRoutes); err != nil {
-		// scale CCM back to originalScale in case if there are errors from DeleteRoutes() call
-		_ = scaleCCMto(originalScale)
-	}
+	for _, route := range cloudRoutes {
```
kon-angelo, line 167: This is a comment for the already existing function name. Since we also have a function for the delete called `ensureKubernetesRoutesDeleted`, I would rename this function that is specific to the ipv6 migration.

pkg/controller/infrastructure/infraflow/ensure.go
```diff
@@ -163,12 +164,31 @@ func (fctx *FlowContext) ensureKubernetesRoutesCleanup(ctx context.Context) erro
...
+		if err != nil {
+			return err
+		}
 
-	return err
+		fctx.state.Routes = append(fctx.state.Routes, apisgcp.Route{InstanceName: instanceName, DestinationCIDR: route.DestRange, Zone: zone})
+		// Delete the route
+		if err := fctx.computeClient.DeleteRoute(ctx, route.Name); err != nil {
+			// scale CCM back to originalScale in case if there are errors from DeleteRoutes() call
+			_ = scaleCCMto(originalScale)
```
kon-angelo, line 177: Is scaling the ccm back to the original scale okay? Does it not need additional options or maybe even allowing the CP controller to run and update the CCM options instead ?

pkg/controller/infrastructure/infraflow/ensure.go
```diff
@@ -163,12 +164,31 @@ func (fctx *FlowContext) ensureKubernetesRoutesCleanup(ctx context.Context) erro
...
+			// if route couldn't be remove last route from stack
```
kon-angelo, line 178: Comment is confusing 😅

pkg/controller/infrastructure/infraflow/ensure.go
```diff
@@ -612,3 +632,34 @@ func (fctx *FlowContext) ensureKubernetesRoutesDeleted(ctx context.Context) erro
 
 	return nil
 }
+
+func (fctx *FlowContext) ensureAliasIpRanges(ctx context.Context) error {
+	log := shared.LogFromContext(ctx)
+
+	for i, route := range fctx.state.Routes {
+		// if machine is already gone procceed with next route
+		if err := fctx.computeClient.InsertAliasIPRoute(ctx, route, DefaultSecondarySubnetName); err != nil && !strings.Contains(err.Error(), "notFound") {
```
kon-angelo, line 641: Can we somehow handle the case where the machine already has an aliasIP much less the same in this or the client function ? I suspect this will come up one way or the other

pkg/controller/infrastructure/infraflow/ensure.go
```diff
@@ -612,3 +632,34 @@ func (fctx *FlowContext) ensureKubernetesRoutesDeleted(ctx context.Context) erro
...
+			perr := fctx.persistState(ctx)
+			if perr != nil {
+				fctx.log.Error(perr, "failed to persist state")
+			}
+			return err
+		} else {
+			log.Info("Successfully added alias IP", "route", route.DestinationCIDR, "instance", route.InstanceName)
+			if i >= 0 && i < len(fctx.state.Routes) {
+				if i < len(fctx.state.Routes)-1 {
+					fctx.state.Routes = append(fctx.state.Routes[:i], fctx.state.Routes[i+1:]...)
```
kon-angelo, line 652: I will be honest that I need to touch up on the go semantics here since we are not only modifying the slice contents but the actual slice header by changing it's length. If you know about it I would be curious of the effects albeit I would prefer avoiding altogether.

pkg/controller/infrastructure/infraflow/ensure.go
DockToFuture: Adding the same aliasIP again won't return in an error. It will just stay as it is.

pkg/gcp/client/compute.go
DockToFuture: Removed it.

pkg/controller/infrastructure/infraflow/ensure.go
DockToFuture: It's fine as when can not switch to alias routes as long as there are routes.

pkg/controller/infrastructure/infraflow/ensure.go
```diff
@@ -163,12 +166,76 @@ func (fctx *FlowContext) ensureKubernetesRoutesCleanup(ctx context.Context) erro
...
+	for err := range errChan {
+		combinedErr = fmt.Errorf("%v; %w", combinedErr, err)
+	}
+
+	// Scale CCM back to originalScale if there were errors
+	if combinedErr != nil {
+		_ = scaleCCMto(originalScale)
+		return combinedErr
+	}
+	return nil
```
kon-angelo, line 238: if we fail the first time executing this function, won't subsequent executions consider the original CCM scale to be 0 ?

Do we really need the scale up or we are okay with the ccm coming up when the CP reconciles.

pkg/controller/infrastructure/infraflow/shared/whiteboard.go
```diff
@@ -57,6 +59,8 @@ type Whiteboard interface {
 
 	// ImportFromFlatMap reconstructs the hierarchical structure from a flat map containing path-like keys
 	ImportFromFlatMap(data FlatMap)
+	// ImportRoutesFromState imports routes from the state. It is used to import routes from the state into the whiteboard.
+	ImportRoutesFromState(key string, routes []gcp.Route)
```
kon-angelo, line 63: I am not sure why we need this interface method and not use the existing set object. Can you elaborate?

You could pull this out in the reconciler if you wish and the generation increase also does not seem to be useful especially when you just read them from the current state.

pkg/gcp/client/compute.go
```diff
@@ -72,6 +74,8 @@ type ComputeClient interface {
 	ListRoutes(ctx context.Context, opts RouteListOpts) ([]*compute.Route, error)
 	// DeleteRoute deletes the specified route.
 	DeleteRoute(ctx context.Context, id string) error
+	// InsertAliasIPRoute creates an alias IP route on the instance
+	InsertAliasIPRoute(ctx context.Context, route apisgcp.Route, defaultSecondarySubnetName string) error
```
kon-angelo, line 78: I don't really like the use of the apisgcp on the client level. You can use another "input type struct" for the client if you wish to. 
Either way when you call this function (in the single usage I saw), you construct a struct anyway.

pkg/gcp/client/compute.go
```diff
@@ -409,6 +413,34 @@ func (c *computeClient) ListRoutes(ctx context.Context, opts RouteListOpts) ([]*
...
+	aliasIpRange := &compute.AliasIpRange{
+		IpCidrRange:         route.DestinationCIDR,
+		SubnetworkRangeName: defaultSecondarySubnetName,
+	}
+
+	// Append the alias IP range to the existing list
+	instance.NetworkInterfaces[0].AliasIpRanges = append(instance.NetworkInterfaces[0].AliasIpRanges, aliasIpRange)
+
+	var op *compute.Operation
+	op, err = c.service.Instances.UpdateNetworkInterface(c.projectID, route.Zone, route.InstanceName, instance.NetworkInterfaces[0].Name, instance.NetworkInterfaces[0]).Context(ctx).Do()
```
kon-angelo, line 434: You sure you don't want to wait for the operation to be completed. This currently is just a "promise" but you don't really wait for the async operation to complete before proceeding.

pkg/controller/infrastructure/infraflow/ensure.go
DockToFuture: Once we have deleted a route the node won't receive any more traffic until the a new route was created by ccm. So I think it is good to scale ccm up. Or how long will it take until a new CP reconcile is happening? Shall we scale it to one if originalScale is 0?

kon-angelo: Won't it start receiving traffic as soon as the alias IP is in place ?

DockToFuture: Yes but we can only attach an aliasIPs to the NIC when all routes were deleted.

kon-angelo: Then maybe I am misunderstanding something here - please entertain me 😅. 

We scale CCM down, delete all routes and add the secondary range to the subnet. So far, so good.

Now let's assume we scale the CCM up. Questions:
- Can we now create routes on top of the alias IP without issues ?
- If yes, then what happens after the CP reconciliation. In the dual-stack case if I remember correctly we disable the route creation. So we will left over uncleaned routes from enabling the CCM here that would not be managed by the CCM anymore.
   - Is my hypothesis so far correct ?
   - Would the existing routes cause issues as nodes roll out ?

Tldr; we migrate the infrastructure (subnet in this case) but the CCM would be scaled up with the old settings. What can we expect to go wrong there ?

DockToFuture: We only scale ccm up when not all routes where deleted in which case we have no alias IPs. Once we have alias IPs we have 0 routes and won't reach this part of code anymore as we are directly return nil if there are 0 routes.

kon-angelo: Ah okay that part is clear now thank you.

Though we lose all progress of the migration in case of an error. Say for example that because of a huge cluster with hundreds of nodes, we hit the context timeout. The subsequent retries will be an arms race with the CCM.

DockToFuture: As we delete the routes in parallel we are really fast, in my test cases the deletion of all routes took not more than 10 to 12 seconds. I've tried to provoke rate limiting or throttling on gcp side with multiple bulk deletions with 200 routes. I never experienced any of this. If you have a better idea how to do this let me know.

kon-angelo (approved):
/lgtm

PR #1012, opened 2025-03-19
deprecate terraformer support

kon-angelo:

/area control-plane
/kind change
/platform gcp

Fixes #

**Release note**:

```breaking operator
The native-SDK infrastructure reconciler is now promoted as default. Existing shoots will be migrated during their maintenance period and new shoots will automatically use this implementation without any need for the added annotations. Migration from the Terraformer implementation is still supported but will be removed in subsequent releases.
```


pkg/controller/infrastructure/actuator_reconcile.go
```diff
@@ -6,34 +6,184 @@ package infrastructure
...
+
+	return state, nil
+}
+
+func (a *actuator) migrateFromTerraform(ctx context.Context, log logr.Logger, infra *extensionsv1alpha1.Infrastructure) (*gcp.InfrastructureState, error) {
+	state := &gcp.InfrastructureState{
+		Data: map[string]string{},
+	}
+	// we want to prevent allowing the deletion of infrastructure if there may be still resources in the cloudprovider.
+	// We will initialize the data with a specific "marker" so that the deletion.
```
hebelsan, line 107: This sentence reads unfinished.

pkg/controller/infrastructure/actuator_reconcile.go
```diff
@@ -6,34 +6,184 @@ package infrastructure
...
 
-func (a *actuator) reconcile(ctx context.Context, logger logr.Logger, selectorFn SelectorFunc, infra *extensionsv1alpha1.Infrastructure, cluster *controller.Cluster) error {
-	useFlow, err := selectorFn(infra, cluster)
+func (a *actuator) reconcile(ctx context.Context, log logr.Logger, infra *extensionsv1alpha1.Infrastructure, cluster *controller.Cluster) error {
+	var (
+		infraState *gcp.InfrastructureState
+		err        error
+	)
+
+	// when the function is called, we may have: a. no state, b. terraform state (migration) or c. flow state. In case of a TF state
```
hebelsan, line 40: I don't really understand this comment.
It seems like there's something missing after `In case of a TF state`

pkg/controller/infrastructure/actuator_reconcile.go
```diff
@@ -6,34 +6,184 @@ package infrastructure
...
-		a:   a,
+	if fsOk {
+		// if it had a flow state, then we just decode it.
+		infraState, err = a.infrastructureStateFromRaw(infra)
+		if err != nil {
+			return err
+		}
+	} else {
+		// otherwise migrate it from the terraform state if needed.
+		infraState, err = a.migrateFromTerraform(ctx, log, infra)
```
hebelsan, line 55: how much longer will we support migrateFromTerraform?

pkg/controller/infrastructure/actuator_delete.go
```diff
@@ -6,40 +6,61 @@ package infrastructure
...
 	if err != nil {
 		return err
 	}
 
-	return reconciler.Delete(ctx, infra, cluster)
+	tf, err := NewTerraformer(log, a.restConfig, "infra", infra, a.disableProjectedTokenMount)
+	if err != nil {
+		return err
+	}
+	return CleanupTerraformerResources(ctx, tf)
```
hebelsan, line 65: Shouldn't we add a comment to remove these line in future?

hebelsan (commented):
Just small comments.
Otherwise looks very good!
Our code base is becoming much leaner :)

pkg/controller/infrastructure/actuator_reconcile.go
kon-angelo: The problem here is that shoots that are hibernated are not reconciled, hence once we remove this line then these shoots may face issues. 

In this case of provider-gcp there migration from terraform is rather minimal and is just a check that indeed there are resources that have been created by terraform. The flow controller is capable of adopting the remaining resources. 

Therefore for provider-gcp, we can expedite this and remove it in 3 releases. Overall, we can keep this code as much as we want since it only relies on the gardener TF library and has nothing to do with the actual terraformer project

hebelsan (approved):
/lgtm

PR #1066, opened 2025-05-15
Fix workerPoolHashDataV2

hebelsan:

/area quality
/kind cleanup
/platform gcp

Currently, the NewWorkerPoolHash does not affect updates to the .spec.provider.workers[].providerConfig field.
This behaviour may change once MCM supports in-place updates for certain operations, such as volume updates.

Fixes #

**Release note**:

```other operator
NONE
```

### DIFF
--- a/pkg/controller/worker/machines.go
+++ b/pkg/controller/worker/machines.go
@@ -11,7 +11,6 @@ import (
 	"path/filepath"
 	"regexp"
 	"slices"
-	"sort"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


kon-angelo (approved):
/lgtm

pkg/controller/worker/machines.go
```diff
@@ -353,55 +352,21 @@ func (w *WorkerDelegate) generateWorkerPoolHash(pool v1alpha1.WorkerPool, worker
 		}
 	}
 
-	workerVolumes := slices.Clone(workerConfig.DataVolumes)
```
acumino, line 356: Shouldn't we add https://github.com/gardener/gardener-extension-provider-aws/blob/39d9b56b5b382e35807b15235fc18e7e0a18b9a1/pkg/controller/worker/machines.go#L400-L402 also for gcp?

hebelsan: Don't we do this here:
https://github.com/gardener/gardener-extension-provider-gcp/blob/8699d04202901efdf57522a7f4759183373f858d/pkg/controller/worker/machines.go#L351

kon-angelo: I guess you refer to this @acumino https://github.com/gardener/gardener-extension-provider-gcp/pull/607#pullrequestreview-1464631092

This was an old discussion where somehow we decided that the core::worker.datavolume.encryption does not affect reality and the actual settings that matter are the encryption settings in the providerConfig since the disks are always encrypted.

@hebelsan Can you add a warning in our docs about that (since that comment in the original PR was not addressed)

acumino: Thank you @kon-angelo for explanation.

PR #1069, opened 2025-05-19
[GEP-31] Adapt extension to support InPlace update

acumino:

/area control-plane delivery open-source scalability
/kind enhancement
/platform gcp

This PR adapts the extension to support in-place updates.
- The worker controller needs to set the appropriate update strategy in the machine deployment
- The changes to worker providerConfig or relevant fields(which affect the worker pool hash calculation) should be denied.
- The providerConfig is skipped from worker pool hash calculation.

> [!NOTE]  
> Changing these fields will cause a change in the calculation of the WorkerPool hash. Currently, machine-controller-manager does not provide an UpdateMachine call to update the data volumes or provider config in-place. `gardener-node-agent` cannot update the provider config in-place either. So we disallow changing these fields if the update strategy is in-place.

Part of https://github.com/gardener/gardener/issues/10219

**Release note**:

```feature user
This extension now supports in-place node updates. Read more about it [here](https://github.com/gardener/gardener/blob/master/docs/proposals/31-inplace-node-update.md).
```
```other operator
Update `gardener/gardener` to `v1.120.0` [Release Notes](https://github.com/gardener/gardener/releases/tag/v1.120.0).
```

### DIFF
--- a/example/20-crd-extensions.gardener.cloud_controlplanes.yaml
+++ b/example/20-crd-extensions.gardener.cloud_controlplanes.yaml
@@ -81,6 +81,9 @@ spec:
                 description: |-
                   Purpose contains the data if a cloud provider needs additional components in order to expose the control plane.
                   This field is immutable.
+
+                  Deprecated: This field will be removed in gardener v1.123.0.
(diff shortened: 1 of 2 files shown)


hebelsan (approved):
/lgtm

PR #1074, opened 2025-05-21
Rm dataVolumes.encrypted from hashv2

hebelsan:

/area documentation
/kind cleanup
/platform gcp

We exclude `.spec.provider.workers[].dataVolumes[].encrypted` from the hash calculation because GCP disks are encrypted by default, and the field does not influence disk encryption behaviour.
Everything related to disk encryption is handled by the providerConfig.

Fixes #

**Release note**:

```other operator
NONE
```


kon-angelo (approved):
/lgtm

PR #1020, opened 2025-03-24
cloud-controller-manager: For the command for Kubernetes 1.32

ialidzhikov:
**How to categorize this PR?**

/area control-plane
/kind bug
/platform gcp

**What this PR does / why we need it**:
Shoot creation for Kubernetes 1.32 fails. The cloud-controller-manager is in `CrashLoopBackOff`:
```
% k -n shoot--it--tmjmp-vh7 get po cloud-controller-manager-d994cc448-9sqp7
NAME                                       READY   STATUS             RESTARTS       AGE
cloud-controller-manager-d994cc448-9sqp7   0/1     CrashLoopBackOff   8 (4m3s ago)   20m
```

The reason is:
```
  Warning  Failed     19m (x5 over 20m)   kubelet                Error: failed to create containerd task: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: exec: "/cloud-controller-manager": stat /cloud-controller-manager: no such file or directory: unknown
```

@kon-angelo found that the binary location for `registry.k8s.io/cloud-provider-gcp/cloud-controller-manager:v32.2.2` has changed from `/cloud-controller-manager` to `/ko-app/cloud-controller-manager`.

We decided to drop the command for K8s 1.32 as the binary is correctly set as entrypoint in the corresponding image.

Verify this by running
```
% docker run registry.k8s.io/cloud-provider-gcp/cloud-controller-manager:v32.2.2 --help
```

**Which issue(s) this PR fixes**:
Follow-up after https://github.com/gardener/gardener-extension-provider-gcp/pull/990

**Special notes for your reviewer**:

**Release note**:

```other operator
NONE
```

PR #1092, opened 2025-06-16
Add Kubernetes 1.33 support

plkokanov:

/area open-source usability
/kind enhancement
/platform gcp
/exp intermediate
/topology garden seed shoot
/merge squash

Part of https://github.com/gardener/gardener/issues/11933

I have successfully validated the functionality as follows:

- [x] Create new clusters with versions < 1.33
- [x] Create new clusters with version = 1.33
- [x] Upgrade old clusters from version 1.32 to version 1.33
- [x] Delete clusters with versions < 1.33
- [x] Delete clusters with version = 1.33

**Release note**:

```other operator
The provider-gcp extension does now support shoot clusters with Kubernetes version 1.33. You should consider the [Kubernetes release notes](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.33.md) before upgrading to 1.33.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,8 @@ Please find more information regarding the extensibility concepts and a detailed
 This extension controller supports the following Kubernetes versions:
 
 | Version         | Support | Conformance test results |
-|-----------------|---------| ------------------------ |
+|-----------------|---------|--------------------------|
+| Kubernetes 1.33 | 1.33.0+ | N/A |
 | Kubernetes 1.32 | 1.32.0+ | [![Gardener v1.32 Conformance Tests](https://testgrid.k8s.io/q/summary/conformance-gardener/Gardener,%20v1.32%20GCE/tests_status?style=svg)](https://testgrid.k8s.io/conformance-gardener#Gardener,%20v1.32%20GCE) |
 | Kubernetes 1.31 | 1.31.0+ | [![Gardener v1.31 Conformance Tests](https://testgrid.k8s.io/q/summary/conformance-gardener/Gardener,%20v1.31%20GCE/tests_status?style=svg)](https://testgrid.k8s.io/conformance-gardener#Gardener,%20v1.31%20GCE) |
 | Kubernetes 1.30 | 1.30.0+ | [![Gardener v1.30 Conformance Tests](https://testgrid.k8s.io/q/summary/conformance-gardener/Gardener,%20v1.30%20GCE/tests_status?style=svg)](https://testgrid.k8s.io/conformance-gardener#Gardener,%20v1.30%20GCE) |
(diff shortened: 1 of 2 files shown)


plkokanov:
I wanted to also bump the version of CCM for Kubernetes 1.32 to `v32.2.5` (I have seen similar changes in previous PRs that add support for new k8s versions). However, it seems like they did a change for the image build of CCM `v32.2.5` to use a Dockerfile which only specifies CMD: https://github.com/kubernetes/cloud-provider-gcp/blob/91253c5aff67e3001f79d2ea4a4ec4dc0701dd1d/Dockerfile#L16

This caused the entrypoint to be defaulted to `ko-runner` and the arguments that are passed in [the cloud-controller-manager pod manifest](https://github.com/gardener/gardener-extension-provider-gcp/blob/b98071f1ebdbba7969875438a99f39b06d2a8c3c/charts/internal/seed-controlplane/charts/cloud-controller-manager/templates/cloud-controller-manager.yaml#L43-L68) to be supplied to `ko-runner` instead of the cloud-controller-manager resulting in errors like:
```
flag provided but not defined: -client-ca-file
Usage of /go-runner:
  -also-stdout
        useful with log-file, log to standard output as well as the log file
  -log-file string
        If non-empty, save stdout to this file
  -redirect-stderr
        treat stderr same as stdout (default true)
```

This has been fixed in https://github.com/kubernetes/cloud-provider-gcp/pull/842 but the fix will only be available for the v33.0.0 release. I asked for a back-port in https://github.com/kubernetes/cloud-provider-gcp/pull/842#issuecomment-2987419314, however, I'm not sure if they will do it.

One way around this is to use CCM v33.2.4 which does not have this problem and wait for a fix.

The other option is to modify https://github.com/gardener/gardener-extension-provider-gcp/blob/b98071f1ebdbba7969875438a99f39b06d2a8c3c/charts/internal/seed-controlplane/charts/cloud-controller-manager/templates/cloud-controller-manager.yaml#L39-L41 so that the `command` is given for both k8s 1.31 and k8s 1.32. In that case we have to make sure that CCM v33.2.5 is used and to watch out that they don't change the behaviour for future patch releases.
@ialidzhikov @kon-angelo wdyt? I saw that there were some changes related to this in https://github.com/gardener/gardener-extension-provider-gcp/pull/1020

On a separate note, there are still no official images for CCM `v33.0.0` :/

kon-angelo:
I think we can live with the changes in v32.2.4 plus, they actually also reverted most changes of the v32.2.5 patch in the current release-v1.32 branch

ialidzhikov (approved):
/lgtm

PR #1095, opened 2025-06-18
Feature: CSI Filestore driver

hebelsan:

/area quality
/kind enhancement
/platform gcp

This PR adds support for enabling the CSI Filestore driver.

Fixes #

**Release note**:

```feature operator
Add support for enabling the CSI Filestore driver
```


charts/internal/seed-controlplane/charts/csi-driver-filestore-controller/templates/csi-driver-filestore-controller.yaml
```diff
@@ -0,0 +1,266 @@
...
+      role: controller
+  strategy:
+    rollingUpdate:
+      maxSurge: 25%
+      maxUnavailable: 25%
+    type: RollingUpdate
+  template:
+    metadata:
+      annotations:
+        checksum/configmap-csi-driver-controller: {{ include (print $.Template.BasePath "/config.yaml") . | sha256sum }}
```
kon-angelo, line 25: Curiously, we don't restart the container if the cloudprovider credentials are changed. Maybe that's a nice addition (also to our normal CSI driver). I don't remember if there is any reason not to do it like that.

PR #1123, opened 2025-07-18
Drop support for Shoots with Kubernetes version <= 1.28

georgibaltiev:

/kind cleanup
/platform gcp

Drop support for Shoots, Seeds and Gardens with Kubernetes version <= 1.28.

Part of https://github.com/gardener/gardener/issues/12409

N/A

**Release note**:

```other operator
`provider-gcp` no longer supports Shoots with Кubernetes version <= 1.28.
```

### DIFF
--- a/pkg/webhook/controlplane/ensurer_test.go
+++ b/pkg/webhook/controlplane/ensurer_test.go
@@ -407,7 +407,7 @@ var _ = Describe("Ensurer", func() {
 				Expect(err).To(Not(HaveOccurred()))
 				Expect(&kubeletConfig).To(Equal(newKubeletConfig))
 			},
-			Entry("kubelet >= 1.30, < 1.31", eContextK8s130, semver.MustParse("1.30.1"), map[string]bool{"InTreePluginGCEUnregister": true}),
+			Entry("kubelet < 1.31", eContextK8s130, semver.MustParse("1.30.1"), map[string]bool{"InTreePluginGCEUnregister": true}),
 			Entry("kubelet >= 1.31", eContextK8s131, semver.MustParse("1.31.1"), map[string]bool{}),
 		)
 	})


pkg/features/features.go
```diff
@@ -11,7 +11,7 @@ import (
 
 const (
 	// DisableGardenerServiceAccountCreation controls whether the gcp provider will create a default service account for VMs managed by MCM.
-	// beta: v1.29.0
+	// beta: v1.29.13
```
ialidzhikov, line 14: You could revert this change. 

The version here is related to the provider-gcp version. It denotes that the corresponding feature gate is beta since provider-gcp@v1.29.0.

pkg/webhook/controlplane/ensurer_test.go
```diff
@@ -118,6 +87,7 @@ var _ = Describe("Ensurer", func() {
 	)
 
 	BeforeEach(func() {
+		format.MaxLength = -1
```
ialidzhikov, line 90: It seems like a leftover from debugging.

```suggestion
```

pkg/webhook/controlplane/ensurer_test.go
```diff
@@ -629,10 +568,8 @@ func checkKubeAPIServerDeployment(dep *appsv1.Deployment, k8sVersion string) {
 	switch {
 	case k8sVersionAtLeast131:
 		Expect(c.Command).NotTo(ContainElement(HavePrefix("--feature-gates")))
-	case k8sVersionAtLeast128:
+	default:
 		Expect(c.Command).To(ContainElement("--feature-gates=InTreePluginGCEUnregister=true"))
-	default: // = 1.27
-		Expect(c.Command).To(ContainElement("--feature-gates=CSIMigrationGCE=true,InTreePluginGCEUnregister=true"))
 	}
```
ialidzhikov, line 573: This could be now replaced with an if else statement. It might be acceptable to have a switch case for 3 conditions. IMO it is not justified for only 2 conditions. A simple if else would do the job.

pkg/webhook/controlplane/ensurer_test.go
```diff
@@ -656,12 +592,9 @@ func checkKubeControllerManagerDeployment(dep *appsv1.Deployment, k8sVersion str
...
 		Expect(c.Command).NotTo(ContainElement(HavePrefix("--feature-gates")))
 		Expect(c.Command).To(ContainElement("--allocate-node-cidrs=false"))
-	case k8sVersionAtLeast128:
+	default:
 		Expect(c.Command).To(ContainElement("--feature-gates=InTreePluginGCEUnregister=true"))
 		Expect(c.Command).To(ContainElement("--allocate-node-cidrs=true"))
-	default: // = 1.27
-		Expect(c.Command).To(ContainElement("--feature-gates=CSIMigrationGCE=true,InTreePluginGCEUnregister=true"))
-		Expect(c.Command).To(ContainElement("--allocate-node-cidrs=true"))
 	}
```
ialidzhikov, line 598: Same as above, we could replace with an if else statement.

pkg/webhook/controlplane/ensurer_test.go
```diff
@@ -686,15 +618,12 @@ func checkKubeSchedulerDeployment(dep *appsv1.Deployment, k8sVersion string) {
 	switch {
 	case k8sVersionAtLeast131:
 		Expect(c.Command).NotTo(ContainElement(HavePrefix("--feature-gates")))
-	case k8sVersionAtLeast128:
+	default:
 		Expect(c.Command).To(ContainElement("--feature-gates=InTreePluginGCEUnregister=true"))
-	default: // = 1.27
-		Expect(c.Command).To(ContainElement("--feature-gates=CSIMigrationGCE=true,InTreePluginGCEUnregister=true"))
 	}
```
ialidzhikov, line 623: Same as above, we could replace with an if else statement.

pkg/webhook/controlplane/ensurer_test.go
```diff
@@ -704,9 +633,7 @@ func checkClusterAutoscalerDeployment(dep *appsv1.Deployment, k8sVersion string)
...
```
ialidzhikov, line 638: Same as above, we could replace with an if else statement.

ialidzhikov (commented):
Thanks for the PR! I am leaving preliminary comments inline.

kon-angelo (dismissed):
/lgtm

pkg/webhook/controlplane/ensurer_test.go
```diff
@@ -467,9 +407,7 @@ var _ = Describe("Ensurer", func() {
 				Expect(err).To(Not(HaveOccurred()))
 				Expect(&kubeletConfig).To(Equal(newKubeletConfig))
 			},
-
-			Entry("kubelet = 1.27", eContextK8s127, semver.MustParse("1.27.1"), map[string]bool{"CSIMigrationGCE": true, "InTreePluginGCEUnregister": true}),
-			Entry("kubelet >= 1.28, < 1.31", eContextK8s128, semver.MustParse("1.28.2"), map[string]bool{"InTreePluginGCEUnregister": true}),
+			Entry("kubelet >= 1.30, < 1.31", eContextK8s130, semver.MustParse("1.30.1"), map[string]bool{"InTreePluginGCEUnregister": true}),
```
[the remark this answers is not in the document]
ialidzhikov: Why do we need to specify lower bound here? The unit under test (`EnsureKubeletConfiguration`) has only if check for K8s version < 1.31: https://github.com/gardener/gardener-extension-provider-gcp/blob/0e36688950e7e37f00f089bce5cf905de366bb8c/pkg/webhook/controlplane/ensurer.go#L360-L373

We are rather interested to test with any version which is < 1.31. 

The file contains other constraints which are don't have a lower bound:
- https://github.com/gardener/gardener-extension-provider-gcp/blob/7bd21dacf5e82ecf0b7a21e9454cbe1e46f99c56/pkg/webhook/controlplane/ensurer_test.go#L319
- https://github.com/gardener/gardener-extension-provider-gcp/blob/7bd21dacf5e82ecf0b7a21e9454cbe1e46f99c56/pkg/webhook/controlplane/ensurer_test.go#L279
- https://github.com/gardener/gardener-extension-provider-gcp/blob/7bd21dacf5e82ecf0b7a21e9454cbe1e46f99c56/pkg/webhook/controlplane/ensurer_test.go#L198
- https://github.com/gardener/gardener-extension-provider-gcp/blob/7bd21dacf5e82ecf0b7a21e9454cbe1e46f99c56/pkg/webhook/controlplane/ensurer_test.go#L122

I rather suggest to remove the lower bound so that we don't have adapt this lower bound every time we drop a Kubernetes version.

```suggestion
			Entry("kubelet < 1.31", eContextK8s130, semver.MustParse("1.30.1"), map[string]bool{"InTreePluginGCEUnregister": true}),
```

PR #1136, opened 2025-07-29
remove obsolete TF resources

kon-angelo:
**How to categorize this PR?**

/area control-plane
/kind enhancement
/platform gcp

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Release note**:

```other operator
Deprecate obsolete role, rolebinding and serviceaccounts from terraformer
```

PR #1176, opened 2025-09-19
Fix RBAC of runtime cluster for terraformer migration

shafeeqes:

/kind enhancement
/platform gcp

The rbac for runtime cluster also needs to be adapted for this migration
Follow up after https://github.com/gardener/gardener-extension-provider-gcp/pull/1136

Fixes #

**Release note**:

```other operator
NONE
```


hebelsan:
Thanks for creating this PR @shafeeqes!
We probably need to fix the runtime RBAC for other providers as well, right?

shafeeqes:
If you have migrations there as well, then yes

PR #1197, opened 2025-10-08
Provide configuration for `google-guest-agent`

axel7born:

/area networking
/kind bug
/platform gcp

Alias IPs are currently not working with Cilium, since the google-guest-agent adds a route for the alias IP range that conflicts with a similar route created by cilium. 
This change adds configuration for the google-guest-agent to not create/ remove this route.
The agent needs to be restarted, which is handled by adding the unit. The new unit definition is created in `/etc/systemd/system`  which overrides the original unit in `/usr/lib/systemd/system`. The original unit can't be changed as the file system is read-only.

Fixes #

Tested with garden-linux and suse-chost.

**Release note**:

```other user
`google-guest-agent` is configured to not add a route for alias IPs now. This fixes dual-stack clusters with Cilium.
```

### DIFF
--- a/pkg/webhook/controlplane/ensurer.go
+++ b/pkg/webhook/controlplane/ensurer.go
@@ -7,6 +7,7 @@ package controlplane
 import (
 	"bytes"
 	"context"
+	_ "embed"
 	"fmt"
 	"regexp"
 	"strconv"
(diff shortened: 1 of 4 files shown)


ScheererJ (approved):
/lgtm

kon-angelo (approved):
/lgtm

PR #1227, opened 2025-11-07
Keep gce-ingress enabled, but scale to 0 after migration from dual-stack to single-stack.

axel7born:

/area  networking
/kind enhancement
/platform gcp

Fixes #

**Release note**:

```other operator
Deployment gce-ingress is scaled to 0, but still enabled after migration from dual-stack to single-stack networking.
```

### DIFF
--- a/pkg/controller/controlplane/valuesprovider.go
+++ b/pkg/controller/controlplane/valuesprovider.go
@@ -445,6 +445,11 @@ func (vp *valuesProvider) getControlPlaneChartValues(
 		return nil, err
 	}
 
+	replicas := 0
+	if isDualstackEnabled(cluster.Shoot.Spec.Networking, nil) {
+		replicas = extensionscontroller.GetControlPlaneReplicas(cluster, scaledDown, 1)
+	}
+
 	return map[string]interface{}{
 		"global": map[string]interface{}{
 			"genericTokenKubeconfigSecretName": extensionscontroller.GenericTokenKubeconfigSecretNameFromCluster(cluster),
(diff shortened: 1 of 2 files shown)


pkg/controller/controlplane/valuesprovider.go
```diff
@@ -453,8 +458,8 @@ func (vp *valuesProvider) getControlPlaneChartValues(
 		gcp.CSIControllerName:          csi,
 		gcp.CSIFilestoreControllerName: csiFilestore,
 		gcp.IngressGCEName: map[string]interface{}{
-			"enabled":  isDualstackEnabled(cluster.Shoot.Spec.Networking, nil),
-			"replicas": extensionscontroller.GetControlPlaneReplicas(cluster, scaledDown, 1),
+			"enabled":  isDualstackEnabled(cluster.Shoot.Spec.Networking, cluster.Shoot.Status.Networking),
```
kon-angelo: I don't think this will work as expected. When the chart is disabled (`!enabled`) then it won't be deployed at all. That includes the replica change.

In this case it should always be enabled in all cases, pretty much like aws-custom-route/ipam and azure-remedy controllers

kon-angelo (changes requested):

pkg/controller/controlplane/valuesprovider.go
axel7born: The chart will not be disabled, when migrating back to single stack. `isDualstackEnabled` will return `true` in this case, as we have still have the two CIDRs in the networking status. Probably, the function Name `isDualstackEnabled` is misleading.

PR #1243, opened 2025-11-25
Fix migration to single stack networking

axel7born:

/area networking
/kind bug
/platform gcp

The migration of shoots from dualstack to single stack contains to issues:
* The node subnet is not correctly set to dualstack after the migration
* New nodes are not configured with alias ips.

Fixes #

**Release note**:

```other operator
A bug in the migration of dual-stack to single-stack clusters has been fixed.
```

### DIFF
--- a/pkg/controller/infrastructure/infraflow/ensure_utils.go
+++ b/pkg/controller/infrastructure/infraflow/ensure_utils.go
@@ -374,21 +374,32 @@ func extractInstanceAndZone(nextHopInstance string) (string, string, error) {
 }
 
 // IPFamiliesFromCIDRs returns a slice of IP families (IPv4 / IPv6) corresponding
-// to the provided CIDRs. The position in the returned slice matches the input.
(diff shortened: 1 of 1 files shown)


pkg/controller/infrastructure/infraflow/ensure_utils.go
```diff
@@ -371,3 +372,23 @@ func extractInstanceAndZone(nextHopInstance string) (string, string, error) {
 
 	return instance, zone, nil
 }
+
+// IPFamiliesFromCIDRs returns a slice of IP families (IPv4 / IPv6) corresponding
+// to the provided CIDRs. The position in the returned slice matches the input.
+func IPFamiliesFromCIDRs(cidrs []string) []gardencorev1beta1.IPFamily {
```
hebelsan: This may produce a slice per CIDR (potentially >2, duplicates, or empty).
Can we support these cases?

PR #1283, opened 2026-01-13
Use `client.Reader` instead of `client.Client` where possible

vpnachev:
/area dev-productivity
/kind enhancement
/platform gcp

Use `client.Reader` instead of `client.Client` where possible.
Read-only clients should use `client.Reader` instead of the read-write `client.Client`.

Fixes #

**Release note**:

```other operator
NONE
```

### DIFF
--- a/pkg/admission/validator/shoot.go
+++ b/pkg/admission/validator/shoot.go
@@ -32,7 +32,9 @@ import (
 )
 
 type shoot struct {
-	client         client.Client
+	// reader is cached read-only client, must be initialized via `mgr.GetClient()`
+	reader client.Reader
+	// apiReader is direct (non-cached) read-only client, must be initialized via `mgr.GetAPIReader()`
 	apiReader      client.Reader
 	decoder        runtime.Decoder
 	lenientDecoder runtime.Decoder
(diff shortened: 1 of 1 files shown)


pkg/gcp/credentialsconfig.go
```diff
@@ -75,7 +75,7 @@ type CredentialsConfig struct {
 }
 
 // GetCredentialsConfigFromSecretReference retrieves the credentials config from the secret with the given secret reference.
-func GetCredentialsConfigFromSecretReference(ctx context.Context, c client.Client, secretRef corev1.SecretReference) (*CredentialsConfig, error) {
+func GetCredentialsConfigFromSecretReference(ctx context.Context, c client.Reader, secretRef corev1.SecretReference) (*CredentialsConfig, error) {
```
hebelsan: Probably missed renaming the parameter from `c` to `r`?

hebelsan (commented):
Thanks for the update @vpnachev!
Just a small nit.

hebelsan (dismissed):
/lgtm

vpnachev:
That is actually very good question, the main driver of the change was that when a client is used for read-only operations, it can be actually `client.Reader` to ensure the client is not used for write/delete operations. I think I will have to check somehow if the change could have negative effect due to lack of cache or see if it is possible to enable cache for the reader client (do not know if this is possible at all).

AndreasBurger (dismissed):

Issue #1029, opened 2025-04-09
L7 Application Load Balancer with IPv6 support


dkoval-py:
**How to categorize this issue?**
/area networking, usability
/kind enhancement
/platform gcp

**What would you like to be added**:

With the recent introduction of [Dual-Stack Support](https://github.com/gardener/gardener-extension-provider-gcp/blob/master/docs/usage/ipv6.md#example-configuration), it is now possible to deploy a Network Passthrough L4 Load Balancer with IPv6 capabilities. Additionally, the updated gce ingress controller enables the deployment of an L7 Application Load Balancer using the native Ingress resource. However, it is currently not possible to automatically configure IPv6 support for the L7 Application Load Balancer via this mechanism.

We would like to request the ability to automatically provision an L7 Application Load Balancer with native IPv6 support.

**Why is this needed**:

Due to security and compliance requirements, we are mandated to adopt L7 Application Load Balancers in order to utilize features such as Cloud Armor. Furthermore, IPv6 traffic handling is also a critical requirement in our environment. Therefore, enabling automatic deployment of dual-stack (IPv4/IPv6) L7 Application Load Balancers is essential for meeting both functional and regulatory expectations.

PR #1320, opened 2026-02-12
update ingress gce

kon-angelo:

/area control-plane
/kind enhancement
/platform gcp

Fixes #

**Release note**:

```feature user
Update ingress-gce to v1.38.2
```
```noteworthy operator
Ingress-gce no longer requires deployment of `BackendConfig` CRDs. In addition, the deployment of the default-http-backend in the shoot is no longer necessary and hence removed.
```


charts/internal/seed-controlplane/charts/ingress-gce/templates/deployment.yaml
```diff
@@ -35,24 +35,25 @@ spec:
...
         - name: glbc
           image: {{ index .Values.images "ingress-gce" }}
-          command:
-            - /glbc
+          args:
             - --config-file-path=/etc/kubernetes/cloudprovider/cloudprovider.conf
             - --healthz-port=8086
             - --logtostderr
             - --sync-period=600s
-            - --default-backend-service=kube-system/default-http-backend
```
ScheererJ: If I remember correctly, the default http backend was used for L7 loadbalancing. By removing it, do we effectively disable L7 loadbalancing capabilities? There are people requesting L7 loadbalancing (see #1029).

ScheererJ (commented):
Thanks for taking the time to look into a new `ingress-gce` release.

PR #1385, opened 2026-04-20
Validate capabilityDefinitions in FindImageInCloudProfile

Roncossek:

/area quality
/kind enhancement
/platform gcp

Adds early validation in `FindImageInCloudProfile` that `capabilityDefinitions` is not empty. An empty slice means the caller skipped `NormalizeCapabilityDefinitions()`, which would cause incorrect or confusing failures downstream.

Fixes #

The sibling function `FindImageInWorkerStatus` handles empty `capabilityDefinitions` via a legacy fallback. `FindImageInCloudProfile` has no such path, so the check makes this requirement explicit.

**Release note**:

```bugfix developer
Validate that `capabilityDefinitions` is not empty in `FindImageInCloudProfile` to fail fast when normalization was skipped.
```

### DIFF
--- a/pkg/apis/gcp/helper/helper.go
+++ b/pkg/apis/gcp/helper/helper.go
@@ -85,6 +85,9 @@ func FindImageInCloudProfile(
 	if cloudProfileConfig == nil {
 		return nil, fmt.Errorf("cloud profile config is nil")
 	}
+	if len(capabilityDefinitions) == 0 {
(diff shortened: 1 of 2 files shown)


Roncossek:
cc @AndreasBurger

AndreasBurger (dismissed):

AndreasBurger:
Good thing the tests block, as I managed to miss the failed one. From looking at the log, it seems the failing tests needs to normalize the capabilities before running, as required by the updated func.
I'm going to discard my previous review for now

AndreasBurger (approved):

PR #1398, opened 2026-05-07
Fix/cp mr healthcheck

hebelsan:

/area monitoring
/kind bug
/platform gcp

This PR updates the `csi-filestore-controller` app label, which previously matched the label used by `disk-csi-driver`. This label conflict caused the `controlplane-seed` GMR health checks to fail.

Fixes https://github.com/gardener/gardener-extension-provider-gcp/issues/1358

**Release note**:

```other operator
Fix controlplane-seed MR failing health checks if filestore is enabled 
```

### DIFF
--- a/pkg/controller/controlplane/valuesprovider.go
+++ b/pkg/controller/controlplane/valuesprovider.go
@@ -780,6 +780,8 @@ func cleanupSeedLegacyCSISnapshotValidation(
 // deleteFilestoreControllerDeploymentWithOldSelector deletes the csi-driver-filestore-controller Deployment
 // if it exists with the old selector (app: csi), so it can be recreated with the new selector
 // (app: csi-filestore) that differentiates it from csi-driver-controller for health checks.
+// Returns an error after deletion to force a requeue, since the apply in the same reconcile cycle
+// would otherwise fail due to the immutable spec.selector field.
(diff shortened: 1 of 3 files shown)


pkg/controller/controlplane/valuesprovider.go
```diff
@@ -771,3 +776,28 @@ func cleanupSeedLegacyCSISnapshotValidation(
 
 	return nil
 }
+
+// deleteFilestoreControllerDeploymentWithOldSelector deletes the csi-driver-filestore-controller Deployment
+// if it exists with the old selector (app: csi), so it can be recreated with the new selector
+// (app: csi-filestore) that differentiates it from csi-driver-controller for health checks.
+// Returns an error after deletion to force a requeue, since the apply in the same reconcile cycle
+// would otherwise fail due to the immutable spec.selector field.
+func deleteFilestoreControllerDeploymentWithOldSelector(
```
kon-angelo: Actually that is probably not the correct place as you always have a race condition. It will at least fail in the first application and will always need to be retried. You can include that in a post-reconcile call preferably so that one reconciliation is enough.

hebelsan: Good point, I've put the deletion after the reconciliation.

kon-angelo (commented):
/lgtm

PR #1399, opened 2026-05-07
Allow configuring MTU for gardener-managed VPCs via Shoot Spec

### LINKED ISSUE #1355: As a user, I want to be able to specify the max MTU for a gardener-managed VPC, such that I don't have to bring my own VPC just for using jumbo frames

/area networking
/kind enhancement
/platform gcp

Expose the setting for max MTU as described in [the docs](https://docs.cloud.google.com/vpc/docs/configure-jumbo-frame-mtu-vpc) in the InfrastructureConfig for creating a gardener-managed VPC.

**Why is this needed**:
It is possible to achieve this today by bringing your own VPC, but this adds additional effort on the stakeholder side. All the lifecycle operations with all the error and exception handling is already implemented on Gardener side. It makes little sense to re-implement this on the stakeholder side just to be able to use jumbo frames.

Side note: some GPU machines seem to have different max MTU recommendations than the rest of the machines in GCP: https://docs.cloud.google.com/compute/docs/gpus/gpu-network-bandwidth#mtu-gpu

voelzmo:
/area networking
/kind api-change
/platform gcp

Adds an optional `mtu` field to `InfrastructureConfig.networks` that allows users to specify the maximum transmission unit (1300–8896 bytes) for Gardener-managed VPC networks. This enables configuring jumbo frames (MTU 8896) without requiring users to bring their own VPC.

The field is immutable after creation because GCP requires all VMs on the network to be stopped before the MTU can be changed.

Fixes #1355

- The field is only allowed for Gardener-managed VPCs (rejected when `networks.vpc` is set).
- When unset (`nil`), GCP applies its default of 1460 bytes.
- Users must keep in mind which mtu settings to apply, as there are different recommendations for different VM types: https://docs.cloud.google.com/compute/docs/gpus/gpu-network-bandwidth#mtu-gpu

**Release note**:
```feature user
The `InfrastructureConfig` API now supports an optional `networks.mtu` field (valid range: 1300–8896) to configure the maximum  transmission unit for Gardener-managed VPC networks.
```

### DIFF
--- a/hack/api-reference/api.md
+++ b/hack/api-reference/api.md
@@ -1360,6 +1360,18 @@ string
 <p>FlowLogs contains the flow log configuration for the subnet.</p>
 </td>
 </tr>
+<tr>
+<td>
+<code>mtu</code></br>
+<em>
+integer
+</em>
+</td>
+<td>
+<em>(Optional)</em>
+<p>MTU is the maximum transmission unit size (in bytes) for the VPC network.<br />Only applicable for Gardener-managed VPCs (i.e., when VPC field is not set).<br />Valid values: 1300 to 8896. If unspecified, GCP defaults to 1460.</p>
+</td>
+</tr>
 
 </tbody>
 </table>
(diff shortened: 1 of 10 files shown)


hebelsan (commented):
Thanks for the contribution!
/lgtm
/approve

PR #1423, opened 2026-06-08
Feat: add backupentry controller and permissions to deployment

hebelsan:

/area quality
/kind enhancement
/platform gcp

This PR enables the backupentry controller in the runtime cluster and adds the necessary RBAC rules

Fixes #

**Release note**:

```other operator
Enables the backupentry controller in the runtime cluster
```


charts/gardener-extension-provider-gcp/templates/rbac-runtime.yaml
wpross, line 23: Not really in focus of this PR, but since it's adding resources to this group:
I think the update verb is not required for any of the resources, so could probably be removed.

wpross (changes requested):

wpross:
I would replace `feat:` with Feat: in the Title as backticks in the title so that they do not end up in the final commit message.

charts/gardener-extension-provider-gcp/templates/rbac-runtime.yaml
wpross: I'll open an issue for this.

wpross (approved):
/lgtm

PR #1428, opened 2026-06-10
Use the correct API version for MutatingAdmissionPolicy

DockToFuture:
`MutatingAdmissionPolicy` graduates from `v1alpha1` → `v1beta1` in Kubernetes 1.34 and to `v1` (GA) in 1.36. Render the chart with the correct API version based on the shoot's Kubernetes version, and align the enable/disable logic accordingly:

- `< 1.34`: `v1alpha1`, requires explicit feature gate + RuntimeConfig opt-in
- `>= 1.34 and < 1.36`: `v1beta1`, beta but disabled by default — requires explicit feature gate + RuntimeConfig opt-in
- `>= 1.36`: `v1`, always enabled (feature gate locked on)

### DIFF
--- a/pkg/controller/controlplane/valuesprovider.go
+++ b/pkg/controller/controlplane/valuesprovider.go
@@ -25,6 +25,7 @@ import (
 	versionutils "github.com/gardener/gardener/pkg/utils/version"
 	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	admissionregistrationv1alpha1 "k8s.io/api/admissionregistration/v1alpha1"
+	admissionregistrationv1beta1 "k8s.io/api/admissionregistration/v1beta1"
 	appsv1 "k8s.io/api/apps/v1"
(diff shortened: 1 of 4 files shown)


pkg/controller/controlplane/valuesprovider.go
```diff
@@ -729,28 +737,62 @@ func isUsingCalico(cluster *extensionscontroller.Cluster) bool {
 		*cluster.Shoot.Spec.Networking.Type == "calico"
 }
 
+// constraintK8sGreaterEqual136 is a version constraint for Kubernetes versions >= 1.36.
```
wpross, line 740: I would add a TODO cleanup comment to use the versionutils version from g/g (like in case of the 134 check) once it is available to not forget it.

wpross (changes requested):

pkg/controller/controlplane/valuesprovider.go
DockToFuture: Added it. :)

wpross (approved):
/lgtm

PR #1432, opened 2026-06-18
Fix: reset MTU to zero for VPC update payload

### LINKED ISSUE #1431: Setting custom node MTU size with Cilium as CNI fails for various reasons

/area networking
/kind bug
/platform gcp

**What happened**:

I tried to deploy a Shoot on GCP with:

```yaml
infrastructureConfig:
    kind: InfrastructureConfig
    networks:
        workers: 10.180.0.0/16
        mtu: 8100
...
networking:
    type: cilium
```

and got one time

```txt
task "Waiting until shoot infrastructure has been reconciled" failed: Error while waiting for Infrastructure shoot--pandora--i569855g/i569855g to become ready: error during reconciliation: Error reconciling infrastructure: 1 error occurred: * failed to "ensure VPC": googleapi: Error 400: Other fields cannot be modified when changing MTU., badRequest
```

and 

```txt
task "Waiting until shoot network plugin has been reconciled" failed: Error while waiting for Network shoot--core--mtu-test/mtu-test to become ready: error during reconciliation: Error reconciling Network: error when generating config values advertised addresses missing in shoot status when retrieving kubernetes service host
```

another time.

Shoot being deployed correctly with Node MTUs being set to `8100` and CNI set to Cilium.

Deploy any shoot configuration on GCP with `infrastructureConfig.networks.mtu` set and `networking.type` to `cilium`.

This relates to #1355 and #1399, since these introduce the feature.

**Environment**:

- Extension version:
- Kubernetes version (use `kubectl version`): 1.35.4
- Cloud provider or hardware configuration: GCP
- Others:

FYI @voelzmo

hebelsan:

/area usability
/kind bug
/platform gcp

This PR does two things:
1. Fix a pointer comparison (avoid unnecessary patches) in VPC update
2. Zero out Mtu on desired before patching — since MTU is immutable, it must never appear in a PATCH regardless

Fixes https://github.com/gardener/gardener-extension-provider-gcp/issues/1431

**Release note**:

```bugfix operator
Fix shoot reconcile error if a custom MTU value is set
```

### DIFF
--- a/pkg/gcp/client/updater.go
+++ b/pkg/gcp/client/updater.go
@@ -52,9 +52,11 @@ func (u *updater) VPC(ctx context.Context, desired, current *compute.Network) (*
 	}
 	// dismiss non-functional changes that may block update
 	desired.Description = ""
+	// MTU is immutable after VPC creation; zero it out so it is omitted from the PATCH payload
+	desired.Mtu = 0
 
 	var modified bool
-	if desired.RoutingConfig != current.RoutingConfig {
+	if !reflect.DeepEqual(desired.RoutingConfig, current.RoutingConfig) {
 		modified = true
 	}
 
--- a/pkg/gcp/client/updater_test.go
+++ b/pkg/gcp/client/updater_test.go
@@ -80,27 +80,5 @@ var _ = Describe("VPC Updater", func() {
 			Expect(err).NotTo(HaveOccurred())
 			Expect(result).To(Equal(patched))
 		})
-
-		It("should not patch VPC when RoutingConfig pointer differs but value is equal", func() {
-			// This is the core regression test: pointer comparison would incorrectly
-			// trigger a PATCH even though nothing changed.
-			desired := &compute.Network{
-				Name:                  "test-vpc",
-				AutoCreateSubnetworks: false,
-				RoutingConfig:         &compute.NetworkRoutingConfig{RoutingMode: "REGIONAL"},
-				Mtu:                   8100,
-			}
-			current := &compute.Network{
-				Name:                  "test-vpc",
-				AutoCreateSubnetworks: false,
-				RoutingConfig:         &compute.NetworkRoutingConfig{RoutingMode: "REGIONAL"},
-				Mtu:                   8100,
-			}
-
-			// PatchNetwork must not be called even though desired and current have different pointers
-			result, err := updater.VPC(ctx, desired, current)
-			Expect(err).NotTo(HaveOccurred())
-			Expect(result).To(Equal(current))
-		})
 	})
 })
--- a/pkg/gcp/client/updater_test.go
+++ b/pkg/gcp/client/updater_test.go
@@ -28,10 +28,6 @@ var _ = Describe("VPC Updater", func() {
 		ctx = context.Background()
 	})
 
-	AfterEach(func() {
-		ctrl.Finish()
-	})
-
 	Context("RoutingConfig", func() {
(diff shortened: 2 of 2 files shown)


pkg/gcp/client/updater.go
```diff
@@ -52,9 +52,11 @@ func (u *updater) VPC(ctx context.Context, desired, current *compute.Network) (*
 	}
 	// dismiss non-functional changes that may block update
 	desired.Description = ""
+	// MTU is immutable after VPC creation; zero it out so it is omitted from the PATCH payload
+	desired.Mtu = 0
 
 	var modified bool
-	if desired.RoutingConfig != current.RoutingConfig {
+	if !reflect.DeepEqual(desired.RoutingConfig, current.RoutingConfig) {
```
wpross, line 59: I would add a test to make sure patching works now with a custom Mtu.

wpross (changes requested):
I guess the issue will be closed when referenced with `Fixes #1431`, but only on of the 2 problems is fixed based on this PR.

pkg/gcp/client/updater.go
hebelsan: I've added unit tests for the VPC update.

Regarding the second error reported in the issue, I believe it's a cascading issue and not related to the gcp-extension itself.
I tested this fix in the local-remote setup, and none of the errors described in the issue were reproduced.

pkg/gcp/client/updater_test.go
```diff
@@ -0,0 +1,106 @@
...
+					Expect(nw.Mtu).To(BeZero())
+					return patched, nil
+				})
+
+			result, err := updater.VPC(ctx, desired, current)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(result).To(Equal(patched))
+		})
+
+		It("should not patch VPC when RoutingConfig pointer differs but value is equal", func() {
```
wpross: seems to be the same as the 1st test.

pkg/gcp/client/updater_test.go
```diff
@@ -0,0 +1,106 @@
+package client_test
+
+import (
+	"context"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"go.uber.org/mock/gomock"
```
wpross, line 8: I guess you're not aware, but would be great if you could adapt it and use the new style based on `fakeclient` in order to comply with https://github.com/gardener/gardener/issues/14572

wpross (changes requested):

pkg/gcp/client/updater_test.go
hebelsan: I check your PR https://github.com/gardener/gardener-extension-provider-openstack/pull/1367 for reference.
The fakeclient pattern applies to controller-runtime Kubernetes client mocks. 
In this test we're only mocking the GCP-specific ComputeClient interface.
There's no Kubernetes client involved. Do you have a different approach in mind for this case?

wpross: ah yeah, you're right 👍
Only thing that could be dropped is the `AfterEach()` block, see
https://github.com/golang/mock/blob/main/gomock/controller.go#L87

wpross (changes requested):

voelzmo:
Great catch and bonus points for adding a test! Thanks!

hebelsan:
> Only thing that could be dropped is the AfterEach()

Nice finding, thanks!

wpross (approved):
/lgtm

PR #1442, opened 2026-07-01
Reintroduce v1.32 support

tobschli:

/area quality
/kind regression
/platform gcp

We removed the support for 1.32 at the same time, as introducing 1.36, which we usually do not do.

Fixes #

**Release note**:

```feature user
This extension now supports shoot clusters with Kubernetes version 1.32 again.
```

### DIFF
--- a/imagevector/images.yaml
+++ b/imagevector/images.yaml
@@ -1,5 +1,14 @@
 images:
+# freeze
 # version-mapping
+# Note: do not use the v32.2.5 image of cloud-controller-manager. Its dockerfile only specifies CMD and does not specify an ENTRYPOINT.
+# This causes the container's entrypoint to be "ko-runner" instead of "/cloud-controller-manager". Due to this it cannot be used with the current
+# chart that is used to deploy the cloud-controller-manager. For more information check
+# https://github.com/gardener/gardener-extension-provider-gcp/pull/1092#issuecomment-2987914997 and
+# https://github.com/gardener/gardener-extension-provider-gcp/pull/1092#issuecomment-2987998478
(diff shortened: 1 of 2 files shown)


imagevector/images.yaml
```diff
@@ -1,5 +1,21 @@
 images:
 # version-mapping
+- name: cloud-controller-manager
```
wpross, line 12: You should add the old comment incl. the `freeze` directive to prevent from updating to the non-functions `v32.2.5` image:

```
# freeze
# version-mapping
# Note: do not use the v32.2.5 image of cloud-controller-manager. Its dockerfile only specifies CMD and does not specify an ENTRYPOINT.
# This causes the container's entrypoint to be "ko-runner" instead of "/cloud-controller-manager". Due to this it cannot be used with the current
# chart that is used to deploy the cloud-controller-manager. For more information check
# https://github.com/gardener/gardener-extension-provider-gcp/pull/1092#issuecomment-2987914997 and
# https://github.com/gardener/gardener-extension-provider-gcp/pull/1092#issuecomment-2987998478
# TODO(plkokanov,AndreasBurger,kon-angelo,hebelsan): Bump the patch version of the cloud-controller-manager after the Dockerfile
# for the v32.2.* releases has been updated to contain both a CMD and ENTRYPOINT ref:
# https://github.com/kubernetes/cloud-provider-gcp/pull/842#issuecomment-2987419314
```

wpross (changes requested):

imagevector/images.yaml
tobschli: Added in [934cace](https://github.com/gardener/gardener-extension-provider-gcp/pull/1442/commits/934cacea2c32e94776b43a42f9e45780d3e8ae0b)

wpross (approved):
/lgtm

PR #1452, opened 2026-07-06
fix: restart ingress-gce pod on cloudprovider credential rotation

### LINKED ISSUE #1451: ingress-gce pod is not restarted on cloudprovider credential rotation
## What happened

During a GCP service account key rotation on a live shoot cluster, the `ingress-gce` pod (running on the seed in the shoot namespace) continued using cached/stale credentials. The `glbc` container failed all GCP Compute API calls with `Invalid JWT Signature`, causing `SyncInstanceGroupsFailed` on LoadBalancer services.

The cloud-controller-manager (CCM) was unaffected because it was restarted during reconciliation.

## Root cause (to be verified)

It appears that in `pkg/controller/controlplane/valuesprovider.go`, the CCM pod template includes a checksum annotation for the `cloudprovider` secret, which triggers a rolling restart when the secret data changes:

```go
"podAnnotations": map[string]interface{}{
    "checksum/secret-" + v1beta1constants.SecretNameCloudProvider: checksums[v1beta1constants.SecretNameCloudProvider],
}
```

The `ingress-gce` deployment maybe does not have an equivalent checksum annotation. Since both components mount the same `cloudprovider` secret, this would mean `ingress-gce` is not restarted on credential rotation while CCM is.

## Expected behavior

`ingress-gce` should be restarted when the `cloudprovider` secret is rotated, similar to CCM.

## Possible fix (to be verified)

Adding a checksum annotation to the `ingress-gce` pod template might resolve this:

```go
"podAnnotations": map[string]interface{}{
    "checksum/secret-" + v1beta1constants.SecretNameCloudProvider: checksums[v1beta1constants.SecretNameCloudProvider],
},
```

This needs verification against the current chart templates and the values provider logic.

xavier-re:
/area control-plane
/kind bug
/platform gcp

The `ingress-gce` pod mounts the shoot's `cloudprovider` secret to authenticate against the GCP Compute API. Unlike the `cloud-controller-manager` and the CSI controllers, its pod template did not carry a checksum annotation over that secret, so the deployment was not rolled when the secret changed.

As a result, when the shoot's GCP service account key was rotated, the `glbc` container kept using its cached credentials and all subsequent Compute API calls failed with `Invalid JWT Signature`, surfacing as `SyncInstanceGroupsFailed` events on Services of type `LoadBalancer`.

This PR adds a `checksum/secret-cloudprovider` annotation to the `ingress-gce` pod template, matching the pattern already used by CCM and the CSI controllers.

Fixes #1451

- No chart changes required — the `ingress-gce` deployment template already renders `.Values.podAnnotations` conditionally.
- Only the `cloudprovider` secret is covered here. The `cloud-provider-config-ingress-gce` configmap is still not checksummed; this matches the current behaviour of the CSI controllers and can be addressed separately.

**Release note**:
```bugfix user
The `ingress-gce` pod is now rolled when the shoot's `cloudprovider` secret changes, so GCP LoadBalancer services no longer fail with `Invalid JWT Signature` after a service account key rotation.
```

### DIFF
--- a/pkg/controller/controlplane/valuesprovider.go
+++ b/pkg/controller/controlplane/valuesprovider.go
@@ -485,6 +485,9 @@ func (vp *valuesProvider) getControlPlaneChartValues(
 			"enabled":             IsDualStackEnabled(cluster.Shoot.Spec.Networking, cluster.Shoot.Status.Networking),
 			"replicas":            replicas,
 			"useWorkloadIdentity": shouldUseWorkloadIdentity(credentialsConfig),
+			"podAnnotations": map[string]interface{}{
+				"checksum/secret-" + v1beta1constants.SecretNameCloudProvider: checksums[v1beta1constants.SecretNameCloudProvider],
+			},
 		},
 	}, nil
 }
--- a/pkg/controller/controlplane/valuesprovider_test.go
+++ b/pkg/controller/controlplane/valuesprovider_test.go
@@ -267,6 +267,9 @@ var _ = Describe("ValuesProvider", func() {
 					"enabled":             false,
 					"replicas":            0,
 					"useWorkloadIdentity": false,
+					"podAnnotations": map[string]interface{}{
+						"checksum/secret-" + v1beta1constants.SecretNameCloudProvider: checksums[v1beta1constants.SecretNameCloudProvider],
+					},
 				},
 			}))
 		})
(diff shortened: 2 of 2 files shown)


kon-angelo (dismissed):
/lgtm

overall there some other things that need to be added like the configmap as podtemplate hash but for the credential issue at hand this PR is enough. I will open a followup though

kon-angelo:
Also @xavier-re ideally you use one of the existing git templates to open PRs because this breaks the release logs as is.

PR #1455, opened 2026-07-08
restart ingress-gce and CSI pods on cloudprovider config change

kon-angelo:
/area control-plane
/kind bug
/platform gcp

Follow-up to #1452. That PR added a `checksum/secret-cloudprovider` annotation to the `ingress-gce` pod template so it is rolled on credential rotation. Three cloud-provider configmaps mounted by control-plane components on the seed remained without a checksum annotation:

- `cloud-provider-config-ingress-gce` — mounted by `ingress-gce`
- `csi-driver-controller-config` — mounted by `csi-driver-controller`
- `csi-filestore-controller-config` — mounted by `csi-driver-filestore-controller`

If any of the input values that drive those configmaps change (project ID, network name, subnetwork name for nodes, zone, node tags), the configmap is updated but the corresponding pod keeps running with the previous configuration until it is restarted for some other reason. This PR closes that gap by adding a `checksum/configmap-<name>` annotation to each of the three pod templates, matching the pattern already used by CCM.

The gardener generic actuator only computes a checksum for one configmap (the CCM `cloud-provider-config`) and passes it in the `checksums` map, so the three additional checksums are computed locally in the values provider over the same input values that render each configmap.

- No chart changes required — all three deployment templates already render `.Values.podAnnotations` conditionally.
- The CSI controller and CSI filestore controller share the same input values (`projectID`, `zone`), so a single checksum is reused for both.
- Unit tests in `valuesprovider_test.go` are extended with the two new expected checksum values and cover all three components across the three control-plane chart scenarios.

**Release note**:
```bugfix user
The `ingress-gce`, `csi-driver-controller` and `csi-driver-filestore-controller` pods are now rolled when their respective cloud-provider configmaps change, so configuration updates are picked up without a manual restart.
```

### DIFF
--- a/pkg/controller/controlplane/valuesprovider.go
+++ b/pkg/controller/controlplane/valuesprovider.go
@@ -17,6 +17,7 @@ import (
 	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
 	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
 	extensionsv1alpha1 "github.com/gardener/gardener/pkg/apis/extensions/v1alpha1"
+	"github.com/gardener/gardener/pkg/utils"
 	"github.com/gardener/gardener/pkg/utils/chart"
 	gutil "github.com/gardener/gardener/pkg/utils/gardener"
 	kutil "github.com/gardener/gardener/pkg/utils/kubernetes"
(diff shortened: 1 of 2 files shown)


kon-angelo:
Closing — opened prematurely without owner approval. Branch remains on my fork for review.

PR #1458, opened 2026-07-11
restart ingress-gce on cloud-provider-config change

kon-angelo:

/area networking
/area control-plane
/kind bug
/platform gcp

Extend the `podAnnotations` of the `ingress-gce` deployment with a checksum annotation over the `cloud-provider-config-ingress-gce` ConfigMap it mounts, so the pods roll when the ConfigMap data changes.

Follow-up to #1452, which only covered the `cloudprovider` secret. The generic actuator only computes a checksum for one ConfigMap (the CCM `cloud-provider-config`), so `cloud-provider-config-ingress-gce` is hashed locally in the valuesprovider and included in the ingress-gce podAnnotations.

Any change to the ConfigMap's data (e.g. network name, subnetwork, node tags) now triggers a rolling restart of the ingress-gce deployment, so the pods pick up the updated configuration instead of continuing to serve with stale content.

- The two CSI controller ConfigMaps (`csi-driver-controller-config`, `csi-filestore-controller-config`) are intentionally not covered here. Their data is limited to `projectID` and `zone`, both of which are immutable for the lifetime of a shoot, and the CSI charts already emit an inline `checksum/configmap-…` annotation via `sha256sum` of the rendered template. Adding a runtime hash for them would also break first-time reconciliation, since those ConfigMaps are created by the same `controlPlaneChart` whose values are being computed.
- The `cloud-provider-config-ingress-gce` ConfigMap belongs to the separate `configChart`, which the generic actuator applies before `GetControlPlaneChartValues` is invoked, so the object is guaranteed to exist by the time we hash it.

**Release note**:
```bugfix operator
The `ingress-gce` deployment now rolls when the mounted `cloud-provider-config-ingress-gce` ConfigMap changes.
```

### DIFF
--- a/pkg/controller/controlplane/valuesprovider.go
+++ b/pkg/controller/controlplane/valuesprovider.go
@@ -17,6 +17,7 @@ import (
 	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
 	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
 	extensionsv1alpha1 "github.com/gardener/gardener/pkg/apis/extensions/v1alpha1"
+	"github.com/gardener/gardener/pkg/utils"
 	"github.com/gardener/gardener/pkg/utils/chart"
 	gutil "github.com/gardener/gardener/pkg/utils/gardener"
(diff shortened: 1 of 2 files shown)


hebelsan (commented):
/lgtm

PR #1475, opened 2026-07-23
Add Support for Machine Preservation

thiyyakat:

/area auto-scaling
/kind enhancement
/platform gcp

This PR adds support for the machine preservation feature (introduced in g/g: v1.147 and MCM: v0.62.0 ) by setting the `AutoPreserveFailedMachineMax` field in the `MachineDeployment.Spec`. The value for this field is derived by distributing `MachineControllerManagerSettings.AutoPreserveFailedMachineMax` defined at the worker pool level, across the zones of the worker pool (similar to `worker.Maximum` and `worker.Minimum`).

References:

g/g PR: https://github.com/gardener/gardener/pull/14767
usage doc: https://github.com/gardener/gardener/blob/master/docs/usage/shoot/shoot_machine_preservation.md

Fixes #

**Release note**:

```other operator
Add support for machine preservation by distributing the value of `worker[].machineControllerManager.autoPreserveFailedMachineMax` across zones.
```

### DIFF
--- a/pkg/controller/worker/machines.go
+++ b/pkg/controller/worker/machines.go
@@ -283,6 +283,11 @@ func (w *WorkerDelegate) generateMachineConfig(ctx context.Context) error {
 				}
 			}
 
+			var preserveMax int32
+			if pool.MachineControllerManagerSettings != nil {
+				preserveMax = ptr.Deref(pool.MachineControllerManagerSettings.AutoPreserveFailedMachineMax, 0)
+			}
+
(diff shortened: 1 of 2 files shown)


matthias-horne:
Since this does not appear to be provider specific, is it possible to do the change centrally in https://github.com/gardener/gardener/blob/master/extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go#L40. We then only need to update the gardener/gardener dependency instead of duplicating the code for each provider.

/hold

PR #1488, opened 2026-08-13
Fix network policy annotation in CSI driver controller service v2

oliver-goetz:

/area control-plane
/kind bug
/platform gcp

The fix in #1487 was upside down 😅  

Fixes #

**Release note**:

```other operator
An erroneous network policy annotation in CSI driver controller service has been fixed.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected service networking configuration so port `3301` is represented as a numeric value, improving compatibility with networking integrations.


hebelsan (commented):
/lgtm
/approve

PR #1493, opened 2026-08-17
Add bring-your-own subnet support

hebelsan:

/area networking
/kind enhancement
/platform gcp

Implements the BYO-subnet / user-managed-egress proposal (#1492). 
When networks.subnetWorkers is set, the extension operates in bring-your-own mode: it references pre-existing user subnets instead of creating the worker subnet, Cloud Router, Cloud NAT, and static firewall rules. Egress becomes the user's responsibility.

Fixes #

The two proposal docs (flexible-network-configuration-proposal.md + -spec.md) are the design reference

**Release note**:

```feature operator
Add bring-your-own subnet support
```

## Summary by CodeRabbit

- **New Features**
  - Added support for user-managed worker and services subnets in single-stack and dual-stack configurations.
  - Added validation for subnet references, CIDR ranges, VPC matching, secondary ranges, and networking immutability.
  - Updated bastion and control-plane networking to use user-managed subnets.
  - Added appropriate cleanup for cluster-managed routes and firewall rules.

- **Documentation**
  - Added configuration, IAM, firewall, dual-stack, runtime, and cleanup guidance.

- **Tests**
  - Expanded infrastructure and bastion integration coverage.
  - Extended infrastructure test execution timeouts.


test/integration/infrastructure/infrastructure_test.go
```diff
@@ -974,3 +992,352 @@ func newProviderConfigForExistingVPC() (string, *gcpv1alpha1.InfrastructureConfi
...
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: gcpv1alpha1.SchemeGroupVersion.String(),
+			Kind:       "InfrastructureConfig",
+		},
+		Networks: gcpv1alpha1.NetworkConfig{
+			VPC: &gcpv1alpha1.VPC{Name: networkName},
+			SubnetWorkers: &gcpv1alpha1.SubnetReference{
+				Name:                  workerSubnetName,
+				PodSecondaryRangeName: new(byoPodRangeName),
+			},
```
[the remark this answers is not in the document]
hebelsan: it does compile and new(byoPodRangeName) returns a *string pointing to "pods".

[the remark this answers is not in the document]
hebelsan: the `new` keyword was updated in Go 1.26 you don't seem to have the latest info

docs/usage/ipv6.md
```diff
@@ -158,3 +158,38 @@ The required annotation `cloud.google.com/l4-rbs: enabled` for ingress-gce is ad
...
+```
+
+### Services subnet (`subnetServices`)
+
+A separate services subnet is required for dual-stack shoots.
+The extension uses the `/64` IPv6 prefix GCP assigns to this subnet to derive the IPv6 services CIDR, slicing a `/108` out of it.
+
+- **`stackType: IPV4_IPV6`** and **`ipv6AccessType: EXTERNAL`** — required so GCP assigns the external IPv6 prefix.
+- **Primary IPv4 range** — required by GCP; choose any small non-overlapping range. It is not used by Gardener.
+- No secondary range is needed.
```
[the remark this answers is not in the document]
hebelsan: adjusted the code, so the docu is valid.

pkg/controller/infrastructure/infraflow/graph.go
```diff
@@ -78,11 +109,19 @@ func (fctx *FlowContext) buildDeleteGraph() *flow.Graph {
 	fctx.BasicFlowContext = shared.NewBasicFlowContext().WithLogger(fctx.log).WithSpan()
 	g := flow.NewGraph("infrastructure deletion")
 
+	if fctx.config.IsUserManagedEgress() {
+		return g
+	}
```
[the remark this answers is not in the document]
hebelsan: Added ensureKubernetesRoutesDeleted  but the firewall rules are not created by the reconciler in the byo mode so they can have different names or be absent - in short we don't manage them.

pkg/controller/infrastructure/infraflow/graph.go
```diff
@@ -78,11 +109,25 @@ func (fctx *FlowContext) buildDeleteGraph() *flow.Graph {
...
 
+	if fctx.config.IsUserManagedEgress() {
+		fctx.AddTask(g, "destroy kubernetes routes", fctx.ensureKubernetesRoutesDeleted,
+			shared.Timeout(defaultDeleteTimeout),
+		)
+		fctx.AddTask(g, "destroy CCM firewall rules", fctx.ensureCCMFirewallRulesDeleted,
+			shared.Timeout(defaultDeleteTimeout),
+		)
+		return g
+	}
```
kon-angelo, line 120: I think that this whole deletion flow is skipped because we don't set the "resources exist" marker in reconciliation

pkg/apis/gcp/v1alpha1/types_infrastructure.go
```diff
@@ -44,6 +44,26 @@ type NetworkConfig struct {
...
+	// SubnetWorkers is a reference to a user-managed subnet for worker nodes.
+	// When set, the extension operates in BYO (bring-your-own) subnet mode:
+	// Gardener does not create or delete the subnet; it only attaches to it.
+	// VPC.Name must be set; CloudNAT, Internal, Worker/Workers, FlowLogs, and MTU must not be set.
+	// +optional
+	SubnetWorkers *SubnetReference `json:"subnetWorkers,omitempty"`
+	// SubnetServices is a reference to a user-managed subnet for services (required for dual-stack BYO shoots).
+	// Only valid when SubnetWorkers is set.
+	// +optional
+	SubnetServices *SubnetReference `json:"subnetServices,omitempty"`
```
kon-angelo, line 56: Can these be the same network ?

pkg/controller/infrastructure/infraflow/ensure_utils.go
```diff
@@ -407,3 +409,40 @@ func IPFamiliesFromCIDRs(cidrs []string) []gardencorev1beta1.IPFamily {
 	}
 	return result
 }
+
+func validateWorkerSubnetCIDRRelationships(workerSubnetCIDR string, networking *gardencorev1beta1.Networking) error {
```
kon-angelo: I think this belongs to the configvalidator

kon-angelo (changes requested):
- I feel like there are missign checks for the configvalidator. For example we can precheck the ipv4/ipv6 stack type of the subnet .
- Single stack to dual stack migration is generally allowed and I think it's something that maybe we should forbid for now.

pkg/controller/infrastructure/infraflow/graph.go
hebelsan: Fixed by setting `fctx.whiteboard.Set(CreatedResourcesExistKey, "true")` in `ensureUserManagedWorkersSubnet`

pkg/controller/infrastructure/infraflow/ensure_utils.go
hebelsan: moved to configvalidator

pkg/apis/gcp/v1alpha1/types_infrastructure.go
hebelsan: No, I added validation to enforce this.

hebelsan: In dual-stack BYO mode, the nodes/pods IPv6 range is the workers subnet's /64, and the services IPv6 range is a /108 carved from the services subnet's /64. 
Each GCP subnet has exactly one /64 and the carve always takes the first /108, so if both pointed at the same subnet, the services /108 would sit inside the nodes/pods /64, fully overlapping.

